PR #126, opened 2026-05-23
fix(tabs): skip disabled tabs during keyboard navigation

stevekinney:
## Summary

`<Tabs>` now skips disabled tabs during keyboard navigation, matching the [WAI-ARIA tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/). Previously, arrow keys (and Home/End) walked the full registry, so under the default `activateOnFocus` for horizontal tablists, arrowing onto a disabled tab activated it — the disabled `<TabPanel>` rendered while no tab visually indicated active state. See `docs/design-debt.md` #16.

Highlights:

- Disabled tabs are filtered out of arrow/Home/End navigation. Home and End bypass adjacency and land directly on the first/last enabled tab.
- Navigation anchors on the user's focused tab (with a fallback to the controlled `value`), so a controlled value pointing at a disabled tab does not seed movement from an invisible position.
- The `TabsContext` exposes a new `setDisabled(value, disabled)` method. `<Tab>` calls it from a dedicated `$effect` that has no dependency on `disabled` in the registration effect, so toggling `disabled` at runtime never re-runs the lifecycle effect — which is what kept the parent's `Map` insertion order (i.e., the navigation order) stable.
- Includes a regression test for the runtime disable-then-re-enable toggle, plus tests for middle/leading/trailing skip, Home/End skip, controlled-disabled-value anchoring, and vertical-orientation + `activateOnFocus=false` + disabled.

## Notable engineering decisions

- **`untrack(() => value)` in the registration effect is load-bearing.** Removing it was attempted in response to a committee finding; the runtime-toggle regression test then failed empirically because Svelte 5 re-runs the effect on the `value` prop's slot when a sibling prop on the same render pass triggers a re-evaluation. The `untrack` pattern documents a value-stability constraint: a `<Tab>`'s `value` is its identity key for its lifetime; consumers who need to change it should remount via a different `{#each}` key.
- **Native `disabled` attribute is preserved (no `aria-disabled` switch).** Native disabled buttons cannot receive focus, which is why the Enter/Space disabled-value gate is defense-in-depth that cannot be exercised through realistic interaction. Adopting `aria-disabled` (so AT users can discover disabled tabs via Tab key) is a separate task.
- **Changeset bumped to `minor`** because `TabsContext` is an exported type and `setDisabled` is an additive public API.

## Review committee

Pre-reviewed by:

- typescript-expert
- testing-expert
- frontend-architect
- simplicity-engineer
- junior-engineer
- Codex (gpt-5.4, xhigh reasoning round 1 + high reasoning round 2) — 2 rounds

2 rounds of review. All must-fix items addressed; one finding (`untrack`) was overridden based on empirical Svelte 5 reactivity behavior, with the rationale recorded in `tmp/committee-state.md`.

## Test plan

- [x] `bun run --filter=cinder typecheck` — 0 errors
- [x] `bun run --filter=cinder test` — 2383 pass, 1 skip, 0 fail
- [x] `bun run lint` — clean
- [x] Pre-commit + pre-push hooks all green
- [ ] Reviewer: verify in the playground (`bun run dev`) that the Tabs `with-keyboard` example walks Design → Develop → Review → wraps to Design, never landing on or rendering "Ship"

---

> [!NOTE]
> **Medium Risk**
> Changes core keyboard navigation behavior for `Tabs` and adds a new context method (`setDisabled`), which could affect existing integrations and edge cases around controlled values and runtime prop changes.
> 
> **Overview**
> Updates `Tabs` keyboard handling to **skip disabled tabs** for Arrow/Home/End navigation (and to ignore disabled tabs for Enter/Space manual activation), aligning behavior with the WAI-ARIA tabs pattern.
> 
> Adds `TabsContext.setDisabled(value, disabled)` and tracks disabled state separately from the tab registry so toggling `disabled` at runtime does **not** perturb navigation order; `Tab` now registers once using `untrack(() => value)` and syncs disabled state via a dedicated effect.
> 
> Expands unit coverage with new tests for disabled-skip behavior (including wraparound, Home/End, controlled disabled values, vertical/manual activation) plus a regression test ensuring disable→enable preserves navigation order, and includes a `minor` changeset for the additive context API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f2da8b81b2592a72ec42b7512af79fe7617d8dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/tabs-disabled-keyboard-skip.md
+++ b/.changeset/tabs-disabled-keyboard-skip.md
@@ -1,0 +1,5 @@
+---
+'cinder': patch
+---
+
+`<Tabs>` now skips disabled tabs during keyboard navigation, in line with the WAI-ARIA tabs pattern. Arrow keys, Home, and End walk only over enabled tabs; an all-disabled tablist is a no-op. The `TabsContext` exposes a new `setDisabled(value, disabled)` method that `<Tab>` calls via a dedicated effect, so toggling `disabled` at runtime preserves the navigation order rather than re-inserting the tab at the end of the registry.
--- a/.changeset/tabs-disabled-keyboard-skip.md
+++ b/.changeset/tabs-disabled-keyboard-skip.md
@@ -1,5 +1,5 @@
 ---
-'cinder': patch
+'cinder': minor
 ---
 
 `<Tabs>` now skips disabled tabs during keyboard navigation, in line with the WAI-ARIA tabs pattern. Arrow keys, Home, and End walk only over enabled tabs; an all-disabled tablist is a no-op. The `TabsContext` exposes a new `setDisabled(value, disabled)` method that `<Tab>` calls via a dedicated effect, so toggling `disabled` at runtime preserves the navigation order rather than re-inserting the tab at the end of the registry.
--- a/packages/components/src/components/tab/tab.svelte
+++ b/packages/components/src/components/tab/tab.svelte
@@ -41,29 +41,35 @@
 
   let buttonElement: HTMLButtonElement | undefined = $state();
 
-  // Register on mount and re-register if the button element changes; unregister
-  // on unmount so the parent's navigation order stays accurate. This effect
-  // intentionally does NOT read `disabled` — toggling the disabled flag must
-  // not cleanup+re-register, because that would move the tab to the end of the
-  // parent's Map insertion order (which is the navigation order). The
-  // separate effect below syncs disabled state through `setDisabled`.
+  // Register the tab once when its button element becomes defined and
+  // unregister on unmount. `value` is treated as stable for this tab's
+  // lifetime (it is the identity key used by the parent's Map and the
+  // recommended key for keyed `{#each}` blocks); reading it via `untrack`
+  // ensures that a sibling prop change (e.g., `disabled` toggling) on the
+  // same render pass cannot cause Svelte to re-trigger this effect, which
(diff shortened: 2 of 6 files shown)


stevekinney:
Closing as redundant.

PR #120 (merged 36 minutes before this branch's first commit) shipped a more thorough fix for the same WAI-ARIA bug. Its implementation:

- Uses the shared `handleRovingKeydown` / `isRovingKey` utilities (extracted to a reusable `roving-tabindex.ts` module, also used by SegmentedControl).
- Single registry `Map<string, { button, disabled }>` with a reactive `version` counter, instead of this PR's two-collection approach.
- Decouples roving tabindex from selection via a new `isFocusable` derived — when the selected tab becomes disabled, `tabindex="0"` shifts to the first enabled tab so the tablist stays keyboard-reachable.
- 16 regression tests including dynamic disable/re-enable, all-disabled no-op with `defaultPrevented` assertions, tabindex shifts, and Enter+Space gating.
- Updated `tabs.a11y.md` docs to match.

This PR's narrower fix is correct but duplicative. The merged PR is the better implementation and ships an abstraction (`handleRovingKeydown`) that the codebase will reuse. No need to maintain two parallel fixes.

Sorry for the noise — should have fetched main before starting work.

PR #150, opened 2026-05-24
Add headless overlay primitives

stevekinney:
## Summary

- add headless Portal, FocusTrap, and Transition/Presence primitives with schemas, documentation, examples, tests, and package exports
- refactor overlay consumers to share the portal attachment behavior while preserving inherited direction and theme attributes
- keep TagInput schema generation aligned with its supported input and accessibility props

## Review committee

Pre-reviewed by: frontend, TypeScript, testing, simplicity, junior, and Codex reviewers.

- Codex (gpt-5.4, xhigh reasoning) — 1 round
- 1 round of review
- All accepted must-fix and suggestion items addressed

## Validation

- `bun run --filter=cinder components:generate`
- `bun run --filter=cinder components:check`
- `bun run --filter=cinder exports:check`
- `bun --cwd packages/components test --conditions browser --conditions svelte --parallel=1 src/components/focus-trap/focus-trap.test.ts src/components/portal/portal.test.ts src/components/transition/transition.test.ts src/components/transition/transition.types.test.ts src/components/popover/popover.test.ts`
- `bun run --filter=cinder typecheck`
- `bun run --filter=cinder validate`

---

> [!NOTE]
> **Medium Risk**
> Introduces new focus/portal/transition primitives and refactors overlay consumers (e.g. `Popover`) to rely on shared portal/focus behavior, which can affect accessibility, focus restoration, and DOM placement across the app.
> 
> **Overview**
> Adds new **headless overlay primitives**: `FocusTrap` (Tab containment + focus restore), `Portal` (move subtree to a target with inline fallback + inherited `dir`/theme copying), and `Transition`/`Presence` (mount/unmount coordination for CSS presence state or Svelte transitions), including schemas, docs, examples, tests, and public exports.
> 
> Refactors existing overlays to use the new primitives: `Popover` now portals via `createPortalAttachment` (with attribute inheritance preserved) and several dialogs/popovers switch to the new `createFocusTrap`, while legacy `createFocusTrap`/`createFocusOnMount` are removed from `utilities/attachments.ts` and a shared `readOption` helper is introduced.
> 
> Also updates generated docs/types (array props no longer shown as `readonly`), tightens `TagInput` schema typing/unsupported-prop metadata for input/accessibility attributes, and makes a small `Autocomplete` highlight color tweak.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c49a8bad463c21ad3883618a5c418be99b46697. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


stevekinney:
Bugbot follow-up after pushing `056d68e7` and `7d022d23`:

- **Portal disconnects when `disabled` flips to true** — fixed in 056d68e7. The attachment now records a comment-node anchor at the wrapper's original location and `restoreInline()` reinserts the wrapper when `disabled` is true or the target is unresolved, so the content never vanishes from the DOM. Regression test in `portal.test.ts` ("restores the wrapper inline when disabled flips true").
- **Duplicate `readOption` utility** — fixed in 7d022d23. Both files now import from a single `utilities/read-option.ts`.
- **Redundant `createFocusOnMount` + `createFocusTrap` in `link-popover.svelte` / `thread-popover.svelte`** — bugbot misread the diff. Neither file uses `createFocusOnMount`; the import-line change just consolidated remaining imports. The only attachment in use is `createFocusTrap()`, which handles capture, initial focus, and restore on its own. No code change needed.

stevekinney:
Round 2 review feedback addressed in `b0f33fc9`:

**Cursor**
- `image-lightbox.svelte`: dropped `captureRestoreTarget` and `createFocusOnMount` — the new `createFocusTrap` captures, sets initial focus, and restores on its own.
- `portal.utilities.svelte.ts`: `copyInheritedPortalAttributes` now takes captured initial values and clears `dir` / `data-cinder-theme` on re-run when the new source does not provide them.

**Copilot**
- `portal.svelte`: `shouldRenderChildren` treats `resolved.kind === 'unresolved'` as inline-fallback after hydration so children stay visible.
- `focus-trap.utilities.svelte.ts`: `resolveScopedTarget` wraps `root.querySelector` in try/catch so a malformed `initialFocus` / `fallbackFocus` selector returns `null` instead of throwing.
- `transition.schema.ts` / `transition.types.ts`: exposed `transition` and `transitionParameters` on `TransitionSchemaProps` so the generator records them under `metadata.unsupportedProps`.
- `transition/README.md`: documented `transition`/`transitionParameters` plus Presence's `present`/`forceMount`/`onExitComplete`.
- `portal/README.md`: documented HTMLElement form for `target`.
- `focus-trap/README.md`: documented HTMLElement form for `initialFocus`/`fallbackFocus`.
- `presence.example.svelte` + transition example JSON: bind `opacity`/`transform` to `open` so the example visibly demonstrates enter/exit.

Added regression tests: unresolved-selector inline fallback, stale-attribute clearing, invalid-selector focus fallback.

stevekinney:
Bugbot follow-up addressed in latest commit: `isFocusableCandidate` correctly excludes `tabindex=-1` from tab-order discovery, but explicit `initialFocus` / `fallbackFocus` targets now route through a new `isProgrammaticallyFocusable` check that accepts `tabindex=-1` — the standard pattern for a programmatically-focusable container or heading. Regression test in `focus-trap.test.ts` ("allows explicit fallbackFocus targets with tabindex=-1").

stevekinney:
Round 3 bugbot follow-ups addressed in latest commit:

- **Dead `createFocusOnMount` in `attachments.ts`** — removed the function, its options type, and the helper `FOCUSABLE_SELECTOR` constant. All three former consumers migrated to `createFocusTrap` in earlier commits.
- **Presence $effect re-runs exit path** — read `presenceState` / `isMounted` untracked inside the exit branch. The exit branch wrote to `presenceState`, so a tracked read was making the effect re-run once more and invalidating the first scheduled rAF via `exitGeneration`. Behavior unchanged; one rAF + one generation increment of wasted work removed per exit.

stevekinney:
Round 4 bugbot follow-up addressed: `isProgrammaticallyFocusable` now requires the element to actually be focusable (has `tabindex`, is a natively focusable tag, is an anchor with `href`, or is contenteditable). Belt-and-suspenders: every `.focus()` call in the trap routes through a new `tryFocus()` helper that verifies the focus actually landed and falls through to the next strategy when it didn't — so a future categorization gap or a runtime condition (inert ancestor inserted between resolution and call, etc.) can never silently leave focus outside the trap. Regression test in `focus-trap.test.ts` ("falls back when initialFocus points at an unfocusable element").

stevekinney:
Round 5 bugbot follow-up addressed in latest commit: `completeExit` now increments `exitGeneration` so any pending rAF/timeout/transitionend callbacks for the current exit fail their `generation !== exitGeneration` guard. With `forceMount`, a wrapper-level transitionend arriving before the rAF set `requiredElapsed`/`exitStart` would otherwise pass the (0 - 0 < 0) timing check, run `completeExit`, and then the still-pending rAF would re-enter `completeExit` via the inner rAF/setTimeout — double-firing `onExitComplete`. Regression test pins `exitCount === 1` in that scenario.

PR #185, opened 2026-05-28
Differentiate Alert vs Callout per P6-C2: lock Alert's live region, pin border contracts

stevekinney:
## Summary

Closes the verification + hardening follow-up for design decision **P6-C2** (Alert vs Callout visual differentiation). The CSS treatment itself shipped in #175; this PR proves and locks the contract.

**What changed:**

- **Alert's `role="alert"` is now non-overridable.** Previously the component rendered `role="alert" {...rest}`, so a consumer's `role="status"` in spread props would win the cascade and a leaked `aria-live="polite"` would fight the role's implicit `aria-live="assertive"`. The component now scrubs `role`, `aria-live`, `aria-atomic`, and `aria-relevant` from rest and spreads the filtered rest **before** the hardcoded `role="alert"` (mirrors the established `banner.svelte` / `callout.svelte` scrub-then-lock pattern). `role="alert"` implies `aria-atomic="true"` per ARIA 1.2 §6.6.1, so a consumer override would fragment the assertive announcement — hence the wider scrub set. `aria-busy` stays on the surface as a legitimate status flag.
- **`AlertProps` omits the four locked attributes** at the type level (matching the runtime scrub set), so `<Alert role="status" />` is a compile error rather than a silent runtime no-op. A new `alert.type-test.ts` pins this with `@ts-expect-error` regressions, mirroring `callout.type-test.ts`.
- **Relationship-based CSS border-width assertions** (per P6-C2 acceptance, "no magic numbers"): Callout proves `border-inline-start-width > border-inline-end-width` derived from source (so a `1px → 2px` token change won't break the test); Alert proves `start === end` structurally (the only border-affecting declaration in the base rules is the `border` shorthand, which sets every edge equally).

These are CSS **cascade-resolution** tests (postcss source parsing), not browser computed-style — happy-dom does not compute styles from stylesheets and the package ships no browser-test harness for these components.

## Review committee

Pre-reviewed by: typescript-expert, testing-expert, frontend-architect, simplicity-engineer
- Codex (gpt-5.4, xhigh) — unavailable this run (see warning below)
- 2 rounds of review; all must-fix items addressed (type-level `Omit`, atrule-skip guard on the Alert width walk, runtime scrub of `aria-atomic`/`aria-relevant`)
- Round 2: unanimous APPROVED

> [!WARNING]
> Codex (the adversarial second-opinion reviewer) was unavailable for this PR — the review shim exited non-zero (`config profile 'heavy' not found`). See `tmp/committee-state.md` for detail. The four Claude-family subagents reviewed across two rounds.

## Test plan

- Targeted: `TZ=UTC LANG=en_US.UTF-8 bun test --conditions browser --conditions svelte --parallel=1 src/components/alert src/components/callout` → **79 pass, 0 fail**
- Full components suite: `bun run test` → **3582 pass, 1 skip, 0 fail**
- `bun run typecheck` (tsc + svelte-check) → **0 errors**
- `bun run lint` → **0 errors**
- TDD verified: the role/aria-live override tests fail against the pre-fix component.

---

> [!NOTE]
> **Medium Risk**
> Changes accessibility behavior for Alert spread props and public types; mistakes could break assistive-tech announcements, though the change aligns with existing banner/callout patterns and is heavily tested.
> 
> **Overview**
> This PR **hardens design decision P6-C2** (Alert vs Callout differentiation) with tests and accessibility contracts, on top of CSS that already shipped elsewhere.
> 
> **Alert** now treats **`role="alert"` as fixed**: spread props are filtered to drop `role`, `aria-live`, `aria-atomic`, and `aria-relevant`, then the filtered rest is applied **before** a hardcoded `role="alert"` so consumers cannot downgrade to `status` or weaken the assertive live region. **`AlertProps`** uses the same **`Omit`** list at compile time, with **`alert.type-test.ts`** and updated snapshot tests; runtime tests cover type escapes via `as never`.
> 
> **CSS acceptance** adds relationship-based border checks parsed from source (not computed styles): **Alert** must have only the base `border` shorthand among border-affecting rules (no dominant start stripe); **Callout** must have **inline-start width strictly greater than** inline-end width, derived from the stylesheet rather than fixed pixel literals.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 146f54256af3a9f30b8315765300436a094b50ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/components/src/components/alert/alert.css.test.ts
+++ b/packages/components/src/components/alert/alert.css.test.ts
@@ -107,47 +107,22 @@ function findRule(selector: string): Rule {
 
 const VARIANTS = ['info', 'success', 'warning', 'error'];
 
-/**
- * Every property that can set an inline-edge width directly or via a shorthand.
- * The base rule sets width only through the `border` shorthand; if NONE of these
- * appear anywhere in the sheet, both inline edges resolve to the base shorthand
- * width and are therefore equal — which is exactly Alert's stripe-free contract.
- */
-const INLINE_WIDTH_AFFECTING = new Set([
-  'border',
-  'border-width',
-  'border-left',
-  'border-right',
-  'border-left-width',
-  'border-right-width',
-  'border-inline',
-  'border-inline-start',
-  'border-inline-end',
-  'border-inline-width',
-  'border-inline-start-width',
-  'border-inline-end-width',
-]);
-
 describe('alert chrome reduction — border-equals-base', () => {
   test('inline-start width EQUALS inline-end width (no stripe — relationship, not magic number)', () => {
     // P6-C2 acceptance: Alert must NOT have a dominant start edge — start width
-    // must equal end width. Rather than hardcoding 1px === 1px, prove it
-    // structurally: the ONLY rule that sets any inline-edge width is the base
-    // `border` shorthand (which sets both edges to the same value). No other
-    // width-affecting declaration exists anywhere in the sheet, so both inline
-    // edges resolve to that one shorthand width and are necessarily equal. A
-    // future edit that lifts one edge would add a declaration here and fail.
+    // must equal end width. Prove it structurally: walk the base rules for
+    // .cinder-alert and assert the only border-affecting declaration is the
+    // `border` shorthand (which sets every edge to the same value). Any future
+    // edit that adds a width longhand would appear here and fail. Uses the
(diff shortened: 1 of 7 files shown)


stevekinney:
@copilot review

PR #192, opened 2026-05-29
feat(select, combobox): make value type generic over options (T extends string)

stevekinney:
## Summary

Make `Select` and `Combobox` generic over their option value type (`T extends string = string`), inferred from their `options` collection — the same pattern `SegmentedControl` already proves. A consumer binding to an enum union now gets compile-time rejection of out-of-union values; untyped consumers (mutable `string[]` options) are unaffected (`T` defaults to `string`, non-breaking).

**Scope:** ONLY Select & Combobox. Tabs & RadioGroup are byte-for-byte unchanged — they have no `options` collection to narrow against, so a root-only generic would be unsound (a `<Tab value="x">` child isn't union-validated across the Svelte context boundary).

### Changes
- **types**: `SelectProps<T>`, `ComboboxProps<T>`, `SelectOption<T>`, `ComboboxOption<T>`. `options` is `readonly` (sole inference source). `value` and value-bearing callbacks use `NoInfer<T>` so they consume the inferred union but never widen it. Sentinels (Step-0 decision table): Select `value?: NoInfer<T>` (`undefined`), Combobox `value?: NoInfer<T> | ''` (preserves the existing `''` empty value — no runtime change). Combobox `filter` takes `ComboboxOption<NoInfer<T>>`.
- **components**: `select.svelte` / `combobox.svelte` declare `generics="T extends string = string"`, `$props()` typed `Props<T>`; value writes only from matched options/selections (never a DOM/input string cast).
- **schema generator**: detect `NoInfer<T>` as `generic-type-parameter`; prefer the locally-authored prop declaration over an inherited HTML-attribute declaration — scoped so it only substitutes when the local type references `NoInfer` (other components shadowing `value` with plain `string` are unaffected). Verified by a full regen: only select & combobox schemas change.
- **api-contract.ts**: update the frozen select contract for the intentional `value` (required→optional, `TSStringKeyword`→`TSTypeReference`) and `options` (`TSArrayType`→`TSTypeOperator`/readonly) changes.
- **type tests** prove COMPONENT-level inference, not pinned annotations. Real `<Select>`/`<Combobox>` markup for the positive cases, plus an inference-mirroring helper for the inline-invalid negative case. **Verified non-vacuous**: replacing `NoInfer<T>` with plain `T` makes the `@ts-expect-error` report as unused (svelte-check fails), so the tests genuinely guard the NoInfer mechanism rather than passing on a pinned type annotation.
- One typed `as const` playground example each for Select & Combobox.

## Review committee

Pre-reviewed by: typescript-expert (implementation), with hand-verification of the NoInfer non-vacuity and the frozen-contract update.

> [!NOTE]
> Direct PR per maintainer request. Committee-review not run for this one.

## Test plan
- `bunx svelte-check` — 0 errors (118 pre-existing warnings); `@ts-expect-error` directives verified to fire.
- Full components suite — 3672 pass / 2 skip / 0 fail (includes the frozen api-contract test).
- `bun test src/components/select src/components/combobox` — 52 pass.
- Schema regen: only select & combobox change (`value` → `generic-type-parameter`); no other component schema drifts.
- `bun run --filter=cinder lint` — 0 errors.
- Tabs & RadioGroup: `git status` shows zero edits.

---

> [!NOTE]
> **Medium Risk**
> Public TypeScript API and Select optional-value semantics change for form controls; schema-generator resolution is scoped but touches shared prop-type logic.
> 
> **Overview**
> **Select** and **Combobox** are now generic over option value type `T` (default `string`), with `T` inferred from `readonly` `options` and `value` typed as `NoInfer<T>` so inline literals cannot widen the union. **Select** makes `value` optional (`undefined` unselected); **Combobox** keeps `''` as the empty sentinel via `NoInfer<T> | ''`.
> 
> The **schema generator** treats `NoInfer<…>` as a generic prop, prefers local prop declarations over inherited HTML `value`, and updates generated schemas/docs so `value` is opaque for these components only. The frozen **`api-contract`** for **select** reflects optional bindable `value` and readonly `options`.
> 
> **Compile-time and runtime tests** (including inference-widening guards), a **Select** test for missing `value`, **`as const` examples**, and a higher default timeout for promotion-check subprocess tests round out the change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit daf53cb8dd52572d3b6408ff4227d3ffac714b64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/components/scripts/check-promotion-readiness.test.ts
+++ b/packages/components/scripts/check-promotion-readiness.test.ts
@@ -7,7 +7,7 @@
  *   - --json output shape and exit-code contracts (via Bun.spawn)
  */
 
-import { describe, expect, test } from 'bun:test';
+import { describe, expect, setDefaultTimeout, test } from 'bun:test';
 import { existsSync, rmSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
@@ -24,6 +24,14 @@ import {
   isInteractive,
   PROP_NAME_DENYLIST,
 } from './component-conventions.ts';
+
+// Several tests spawn the promotion-check CLI as a full `bun` subprocess (schema
+// regeneration + ts-morph/svelte AST parsing per run). Under the parallel
+// pre-push suite those subprocesses are starved and routinely exceed bun:test's
+// default 5s per-test timeout, producing load-induced flakes. Raise the ceiling
+// so the spawn-based contract tests stay deterministic under load; the fast
+// in-process unit tests in this file are unaffected by the larger ceiling.
+setDefaultTimeout(30_000);
 
 const SCRIPT_FILE = join(fileURLToPath(import.meta.url), '..', 'check-promotion-readiness.ts');
 
--- a/packages/components/scripts/generate-component-schema.ts
+++ b/packages/components/scripts/generate-component-schema.ts
@@ -293,6 +293,13 @@ function convertSingleType(type: Type, depth = 0, expandObjectShapes = false): C
     return { kind: 'unsupported', reason: 'generic-type-parameter' };
   }
 
+  // NoInfer<T> — TypeScript 5.4 substitution type that wraps a type parameter to
+  // prevent inference widening. ts-morph does not flag it as a type parameter, but
+  // its getText() returns "NoInfer<…>", so we detect it by text prefix.
+  if (type.getText().startsWith('NoInfer<')) {
+    return { kind: 'unsupported', reason: 'generic-type-parameter' };
+  }
+
   // Object: model simple structural object types only when explicitly allowed.
   // Unsupported member shapes make the whole prop unsupported instead of emitting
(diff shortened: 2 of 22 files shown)


stevekinney:
@copilot review

stevekinney:
Addressed in daf53cb8 — updated both stale comments in `combobox.type-test.ts` to match the actual `filter?: (option: ComboboxOption<T>, ...)` type.

On the "could theoretically widen T" note: a callback **parameter** is a contra-variant position, which TypeScript never uses as an inference site for the enclosing generic. So `filter` cannot widen `T` regardless of `NoInfer` — the `NoInfer` wrapper there was inert, which is exactly why it was removed (round-1 frontend-architect + typescript-expert both confirmed). The real inference-widening guard lives on the co-variant `value` prop (`NoInfer<T> | ''`), which is unchanged and proven non-vacuous in the type-test.

PR #220, opened 2026-05-31
fix(markdown): kill the main-green EISDIR flake via static import-graph traversal

stevekinney:
## Summary

`packages/markdown/src/import-graph.test.ts` intermittently fails `main-green` on Linux CI with `EISDIR reading file: ".../packages/diff/src/line-diff.ts"`. This rewrites the test to eliminate the race at its root.

**Root cause** (confirmed via a second-opinion analysis): *not* a normal concurrent-read race (POSIX allows that). It's a **Bun bundler / libuv / overlayfs path-confusion** that surfaces when `Bun.build` walks the live workspace graph (reading `line-diff.ts` as the deepest leaf) while `@cinder/diff`'s own `bun test` process touches that same file under `bun run --filter='*' test`. Prior PRs #213/#214 added retries — those only lowered the probability.

**Fix:** replace `Bun.build` with a **static import-graph traversal**. Resolve each subpath to its source `.ts` via the package `exports` map (conditions `bun > browser > import > default`), read with `readFileSync`, extract runtime imports via `Bun.Transpiler.scanImports` (excludes `import type`; includes `export * from` + dynamic imports), resolve + recurse, and fail if a forbidden rendering package (`shiki`/`@shikijs/`/`rehype-katex`/`remark-math`) is reachable. Pure file reads — no bundler, immune to the race, ~18ms vs ~340ms/assertion.

**Stricter, and closes every edge-hiding path** the committee surfaced:
- FAILS LOUDLY on any internal (relative or `@cinder/*`) edge it cannot follow — a dropped edge can't silently hide a dependency.
- BANS computed dynamic imports (`import(variable)`, `import('a'+b)`, interpolated templates, conditionals) in guarded source via a **TypeScript AST** check — those are invisible to static analysis.
- Guarantee made explicit: asserts cinder *source* never *directly* imports a rendering package (does not traverse node_modules); safe because cinder controls its direct deps and the forbidden list is complete.

## Review committee

Pre-reviewed by: testing-expert, typescript-expert, simplicity-engineer
- Codex (gpt-5.4, xhigh reasoning) — **4 rounds**
- All must-fix items addressed

Codex drove the robustness: it caught the silent-unresolved-edge false negative, then that computed imports were documented-but-not-enforced, then a regex bypass for literal-prefixed computed expressions — leading to the final AST-based detector.

## Test plan

- `bun test src/import-graph.test.ts` — 7 pass / 0 fail (incl. computed-import enforcement covering variable / concatenation / interpolated-template / conditional, and must-not-trip cases: `reimport(`, `obj.import()`, `import(` in comments and strings).
- Full `@cinder/markdown` suite: 752 tests / 0 fail. Verified clean across repeated runs **and concurrently with `@cinder/diff`** (the original race scenario).
- typecheck + lint + format clean; pre-commit + pre-push green.

---

> [!NOTE]
> **Low Risk**
> Test-only change in import-graph.test.ts; no production runtime or API behavior changes, only how bundle-bloat regressions are guarded in CI.
> 
> **Overview**
> Replaces the `@cinder/markdown` **import-graph leanness** guard so it no longer bundles subpaths with `Bun.build` (which caused intermittent Linux CI `EISDIR` failures on `line-diff.ts`). The test now **walks the workspace source graph statically**: resolve entries via `package.json` `exports`, read files, collect runtime imports with `Bun.Transpiler.scanImports`, and recurse while flagging direct imports of forbidden rendering packages (`shiki`, `@shikijs/*`, `rehype-katex`, `remark-math`).
> 
> The new approach is **stricter in-repo** than grepping bundle text: unresolved relative or `@cinder/*` edges **fail loudly**, and **computed** `import(...)` / `require(...)` specifiers are **banned** via a TypeScript AST check so dependencies cannot hide from static analysis. Leanness cases stay the same (lean subpaths/root/barrels); the rendering subpath sanity check now asserts those packages are **reachable** in the graph, with added tests pinning `scanImports` behavior and the computed-import ban.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 621e03e2448297eba8b0551a091ba976835df83d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/markdown/src/import-graph.test.ts
+++ b/packages/markdown/src/import-graph.test.ts
@@ -1,277 +1,293 @@
 /**
  * Import-graph leanness guard.
  *
- * The bundle-bloat regression we just fixed (rendering pipeline pulled in
- * by every consumer of @cinder/markdown) is the kind of thing that's
- * easy to re-introduce — one accidental `export *` or a barrel import
- * across subpaths and the rendering stack creeps back into diff-only
- * consumers.
+ * The bundle-bloat regression this protects against (the rendering pipeline
+ * pulled in by every consumer of @cinder/markdown) is easy to re-introduce —
+ * one accidental `export *` or a barrel import across subpaths and the rendering
+ * stack creeps back into diff-only consumers.
  *
- * This test bundles each subpath in isolation via Bun.build and
- * asserts the resulting bundle text does NOT contain any of the heavy
- * rendering-only modules. It does NOT enforce a "no unified" rule on
- * the pipeline barrels — the parser/serializer legitimately depend on
- * unified/remark-parse/remark-stringify, and forcing those out would
- * defeat the purpose of the aggregate.
+ * HOW THIS WORKS (and why it no longer uses Bun.build):
+ *
+ * It used to bundle each subpath with `Bun.build` and grep the output for
+ * rendering-library strings. On Linux CI that intermittently failed with
+ * `EISDIR reading file: ".../packages/diff/src/line-diff.ts"` — a Bun
+ * bundler / libuv / overlayfs path-confusion surfaced as a spurious filesystem
+ * error when `Bun.build` walked the live workspace graph while @cinder/diff's
+ * own test process touched the same file. Retries (PRs #213/#214) only reduced
+ * the probability; they never eliminated it.
+ *
+ * Instead we now walk the import graph STATICALLY: resolve each subpath to its
+ * source file via the package `exports` map, read each file with `readFileSync`,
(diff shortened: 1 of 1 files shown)


stevekinney:
Addressed in 621e03e2: extended assertNoComputedImport to ban computed `require(...)` as well — scanImports reports require-call entries with the same blind spot, so a computed `require(pkg)` is now rejected just like a computed `import(pkg)`. Regression test covers literal/computed/member-call require forms.

Issue #364, opened 2026-06-11
Pre-push test phase races shared-dist rebuilds → non-deterministic "X is not declared in this file"


stevekinney:
## Summary

The scoped pre-push hook (`packages/components/scripts/husky/pre-push.ts`) runs
its validation jobs in three phases (lint → typecheck → test), **parallel within
a phase** via an async pool capped at `min(navigator.hardwareConcurrency, 4)`.

Each package's `test` npm-script runs **inline build steps** for its workspace
dependencies, and those `build` scripts do `rm -rf dist` then re-emit. For example:

```jsonc
// packages/editor/package.json
"test": "bun run --filter='@cinder/diff' build && bun run --filter='@cinder/markdown' build && bun test"
// packages/commentary/package.json
"test": "... && bun run --filter='@cinder/markdown' build && bun run --filter='@cinder/editor' build && bun test"
// packages/markdown/scripts/build.ts
await $`rm -rf dist`;           // ← wholesale delete
const buildResult = await Bun.build({ outdir: distributionDirectory, ... });
```

When the pool runs `@cinder/editor test`, `@cinder/commentary test`, and
`@cinder/playground test` **concurrently**, two of them rebuild the **same**
shared `@cinder/markdown` / `@cinder/editor` `dist/` while `@cinder/playground`'s
page-bundle test bundles/reads that dist mid-rewrite. The reader hits a window
where `dist` is deleted or half-repopulated and fails with:

```
error: "renderMarkdown" is not declared in this file
    at packages/markdown/dist/rendering/index.js:NN:N
```

(27 such lines, all from the single test `serves markdown-backed editor page
bundles and their chunks`.)

This is **non-deterministic** (616-vs-617 pass jitter) and **blocks every push**
whose reverse-dependency closure spans a shared-`dist/` dependency plus ≥2
dependents — i.e. essentially any change under `packages/components` that pulls
markdown/editor + their dependents into the scoped closure.

## Why it's confusing

- The on-disk `dist` after the run is the correct (settled) build, so reading the
  file post-failure shows nothing wrong — you're seeing a transient mid-write state.
- An isolated `bun run --filter=@cinder/playground test` passes (no concurrency).
- `origin/main`'s gating `unit-tests` CI job is green on the same SHA (CI doesn't
  hit this hook), so it reads as a local-only flake.

Two byte-identical failures under different machine load = deterministic, not flake.

## Reproduce (deterministic)

```bash
bun run --filter=@cinder/playground test & \
[...]

PR #365, opened 2026-06-11
fix(hooks): serialize the pre-push test phase to stop shared-dist races

stevekinney:
## Problem

The scoped pre-push hook (`packages/components/scripts/husky/pre-push.ts`) runs validation in three phases (lint → typecheck → test), **parallel within a phase** via an async pool capped at `min(hardwareConcurrency, 4)`.

Each package's `test` npm-script runs **inline build steps** for its workspace dependencies, and those `build` scripts do `rm -rf dist` then re-emit:

```jsonc
// packages/editor/package.json
"test": "bun run --filter='@cinder/diff' build && bun run --filter='@cinder/markdown' build && bun test"
// packages/markdown/scripts/build.ts
await $`rm -rf dist`; const buildResult = await Bun.build({ outdir, ... });
```

When the pool runs `@cinder/editor test`, `@cinder/commentary test`, and `@cinder/playground test` **concurrently**, two of them rebuild the **same** shared `@cinder/markdown` / `@cinder/editor` `dist/` while `@cinder/playground`'s page-bundle test reads it mid-rewrite, producing non-deterministic:

```
error: "renderMarkdown" is not declared in this file
    at packages/markdown/dist/rendering/index.js:NN:N
```

This **blocked every push** whose reverse-dependency closure spans a shared-`dist/` dependency plus ≥2 dependents.

## Fix

Serialize the **test** phase (`runJobs(phaseJobs, 1)`). Lint (oxlint) and typecheck (`tsc --noEmit`) are read-only and stay parallel — no latency change for them. Most pushes touch few packages, so the serialization cost is negligible in the common case; it only appears on wide closures touching base deps.

## Verification

Reproduced deterministically (no `git push` needed):

```bash
bun run --filter=@cinder/playground test & \
bun run --filter=@cinder/editor test & \
bun run --filter=@cinder/commentary test & wait
# → playground exits 1 ("X is not declared in this file")
# Serial → all pass 617/617.
```

- `tsc --noEmit` clean; pre-commit full `@lostgradient/cinder` typecheck + test exit 0.
- Pushing this branch ran the **fixed** hook serially and passed (`Pre-push validation passed (scoped)`) — the end-to-end integration proof.

## Why no unit test in this PR

A call-site regression test ("test phase invoked with concurrency 1") requires importing this gate script, which runs module-scope side effects (`isContinuousIntegration() → process.exit(0)`, stdin read, lock). Guarding those behind `import.meta.main` to make it importable risks a silent-skip false green **on the push gate itself** — the exact failure mode the file's `?? 1` NaN-guard is paranoid about. The honest regression artifact is the deterministic repro above. Extracting the phase runner into the side-effect-free `utilities.ts` for a safe test is filed as a follow-up.

## Follow-ups (issue #364)

1. `runFull` (root `bun run --filter='*' test`) has the same latent race — annotated in this PR, not serialized (it's a fail-safe/escalation path, not routine pushes).
2. **Atomic build output** (temp-dir + rename instead of `rm -rf dist`) is the root fix — would make concurrent builds+reads safe and remove the sum-vs-max latency, fixing (1) too.
3. Phase-runner extraction for a real regression test.

## Committee review

dx-engineer, simplicity-engineer, typescript-expert, and Codex (gpt-5.4) — all APPROVED, zero must-fix. All suggestions addressed (comment de-duplication, `runFull` hazard note; defensive clamp kept per typescript-expert).

---

> [!NOTE]
> **Low Risk**
> Changes only local git hook orchestration in `pre-push.ts`; no runtime app, auth, or data paths.
> 
> **Overview**
> The scoped pre-push hook **runs the test phase one package at a time** so parallel `test` jobs no longer race on shared dependency `dist/` rebuilds (`rm -rf dist` + rewrite), which was causing flaky `"<name>" is not declared in this file` failures and blocking pushes.
> 
> **`runJobs`** now accepts an optional **`maxConcurrency`** (default unchanged: up to 4 workers from `hardwareConcurrency`). Concurrency is **clamped with `Number.isFinite`** so a bad value (e.g. `NaN`) cannot start zero workers and falsely pass every job. Lint and typecheck still use the default parallel pool.
> 
> Comments document why test is serialized and that **`runFull`**’s root `bun run --filter='*' test` can still hit the same race on escalation paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ea02d6a668e250014c26fbc8b7a45f402c1e27d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/components/scripts/husky/pre-push.ts
+++ b/packages/components/scripts/husky/pre-push.ts
@@ -124,14 +124,8 @@ type BatchOutcome = {
  * and return a {@link BatchOutcome} so the caller can render a digest and write
  * the full log.
  *
- * `maxConcurrency` caps the pool. The default (CPU-bound, up to 4) is right for
- * lint/typecheck, which only read sources. The `test` phase must pass `1`: each
- * package's `test` script runs inline `bun run --filter=<dep> build` steps that
- * `rm -rf dist` and re-emit a shared dependency's `dist/` (e.g. `@cinder/markdown`,
- * `@cinder/editor`). Run in parallel, two jobs rebuild the same `dist/` while a
- * third job's bundler reads it mid-write, yielding non-deterministic
- * "X is not declared in this file" errors. Serializing the test phase removes the
- * shared-`dist/` write hazard.
+ * `maxConcurrency` caps the pool; the default is CPU-bound (up to 4). The test
+ * phase passes `1` to serialize — see the call site in `runScoped` for why.
  */
 async function runJobs(
   jobs: readonly Job[],
@@ -242,6 +236,11 @@ async function runFull(): Promise<boolean> {
     // The full-suite path can produce very large output. Stream it directly so
     // Bun never keeps the hook alive waiting on captured pipe readers after the
     // command has exited.
+    // NOTE: the root `test` script is `bun run --filter='*' test`, which runs
+    // every package's tests concurrently — so the same shared-`dist/` rebuild
+    // race the scoped path serializes around (see `runScoped`) can still occur
+    // here. The atomic-build-output follow-up is the real fix; this path is
+    // taken only on root-config escalation / fail-safe, not routine pushes.
     const result = await runHookCommand('bun', ['run', script], {
       cwd: REPO_ROOT,
       stderr: 'inherit',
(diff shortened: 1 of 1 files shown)


stevekinney:
Good catch — fixed in 1ea02d6a. You're right that `Math.max(1, maxConcurrency)` doesn't neutralize a non-finite value: `Math.max(1, NaN)` is `NaN`, which would make `Math.min(NaN, jobs.length)` NaN, start zero workers, and silently pass every job. Replaced it with `Number.isFinite(maxConcurrency) ? Math.max(1, Math.floor(maxConcurrency)) : 1` so NaN/±Infinity/0/negatives all floor to 1 (never zero workers), and corrected the stale comment that claimed the clamp prevented the NaN case.

PR #371, opened 2026-06-14
fix(#364): serialize the pre-push test gate (+ atomic-swap defense-in-depth)

### LINKED ISSUE #364: Pre-push test phase races shared-dist rebuilds → non-deterministic "X is not declared in this file"

stevekinney:
Closes #364.

## The bug

The husky pre-push (and pre-commit) test gate runs sibling package `test` scripts in parallel. Several of those scripts (`commentary`, `editor`, `playground`) rebuild upstream packages inline via `bun run --filter='@cinder/diff' build` (plus `markdown`/`editor`/`commentary`). The old build scripts did `rm -rf dist` and then wrote into `dist/` in place, so a consumer reading `dist/...` while a sibling was mid-rebuild could observe a **partial or missing** tree — surfacing as the non-deterministic `"X is not declared in this file"` failures in #364.

## The fix (two layered defenses)

**1. Serialize the test-gate phase — this is the actual #364 fix.** A `phaseMaxConcurrency` seam (`412440ca`) makes the test phase run with `maxConcurrency === 1`. With the test scripts serialized, no two of them rebuild a shared `dist/` concurrently inside the hook, so the race that produced #364 cannot occur in the gate.

**2. Per-PID staging + atomic swap — defense-in-depth for out-of-hook builds.** `atomicSwapDist` (`5574a85d`, refined in round 1) builds each upstream package into a per-PID staging directory (`dist.tmp-<pid>-<time>-<rand>`) and atomically `rename`s it into place, instead of mutating `dist/` in situ. This protects the **turbo / CI / manual** build paths that the husky serialization does **not** govern — e.g. a developer's `bun run build` racing CI's build of the same package. It is explicitly **not** a replacement for the serialization: the rebuild path still has a documented sub-millisecond transient-ENOENT window that atomic swap does not close. The two defenses cover different execution paths; neither makes the other redundant.

## Scope audit (every package build script accounted for)

A repo-wide audit confirms the change is scoped to exactly the race victims:

- **`diff`, `markdown`, `editor`, `commentary`** — race victims. Their `dist/` is rebuilt by sibling test scripts via inline `bun run --filter=… build`. These four get the staging-dir + atomic-swap treatment.
- **`components` (`@lostgradient/cinder`)** — **not** a victim. No sibling test script rebuilds it during the gate, so its `build.ts` `rm -rf dist` is intentionally left unchanged. Adding staging/swap here would be unnecessary churn.
- **`playground`, `testing`** — no `dist` build to race.

## Round-2 addition: prove the runner *honors* the cap

The `phaseMaxConcurrency` policy tests proved the test phase *asks* for concurrency 1, but nothing proved the runner *honors* it — a future edit hardcoding `4` into a `runJobs(…)` call would leave every existing test green. To close that policy/mechanism gap:

- Extracted the bounded async pool both hooks' `runJobs` share into `runWithConcurrencyPool<Item, Result>(items, maxConcurrency, worker)` in `utilities.ts`, placed next to `phaseMaxConcurrency` as its policy/mechanism pair. The non-finite-cap floor (`Number.isFinite ? Math.max(1, floor) : 1` — NaN false-green protection) now lives in one place.
- `pre-commit.ts` and `pre-push.ts` both delegate to it. Pre-push keeps its per-job spawn-failure catch and `BatchOutcome` digest shaping; the early-failure/abort path and return type are unchanged — this is the pool primitive only, not a merge of the two runners.
- Added a 6-test `runWithConcurrencyPool` block that instruments the worker to record max overlap: `maxConcurrency === 1` → strictly 1 (the #364 guarantee), `> 1` → overlaps up to the cap, never more workers than items, empty-input no-op, NaN floors to 1, and input-order preservation when later items resolve first.

## On the duplicated helper

`atomic-swap-dist.ts` is duplicated byte-for-byte across the five package `scripts/lib/` directories rather than shared from a repo-root module. This is deliberate: each package's `tsconfig.json` sets `"rootDir": "."`, so a `scripts/build.ts` importing a repo-root shared module puts a source file above `rootDir` → `TS6059` (verified empirically). Lifting the helper to a shared module would require per-package `rootDir`/project-reference changes across **all five** tsconfigs to avoid touching five source files — strictly more moving parts. Drift is mechanically guarded by `atomic-swap-dist.duplication.test.ts`, a byte-equality check across all five copies (verified to catch drift via a negative control).

## Testing

- `utilities.test.ts` (policy + new `runWithConcurrencyPool` mechanism): 106 pass / 0 fail.
- `atomic-swap-dist` suites (real-FS happy/unexpected-error, injected-fs deterministic step-2/step-3/re-throw, concurrency regression with negative+positive controls, duplication guard): 45 pass / 0 fail.
- Full components pre-commit suite green; pre-push validation passed 7/7 packages.

## Review

Opened after a committee review (Codex `xhigh` primary + per-persona testing/simplicity/typescript/junior/dx fallbacks at `medium`). Round-1 must-fixes addressed (narrowed errno handling, unique PID+time+random suffixes, corrected serialization-vs-swap comments, injectable-fs test seam, concurrency regression + duplication guards). Round-2 re-review reached consensus: primary Codex, testing-expert, and simplicity-engineer all **APPROVED**.

The two round-1 simplicity-engineer must-fixes (shared module; pick one defense) were adjudicated and rejected on the merits — empirical `TS6059` proof for the duplication, and a cross-model design adjudication for keeping both defenses (they cover the hook path vs. the out-of-hook path respectively). Both were upheld by simplicity-engineer's round-2 re-review.

/cc @copilot — review requested.

---

> [!NOTE]
> **Medium Risk**
> Touches developer workflow (pre-commit/pre-push timing) and all four upstream package build scripts; incorrect concurrency or swap logic could mask real failures or leave packages without `dist`, though extensive regression tests mitigate this.
> 
> **Overview**
> Fixes **#364** by stopping parallel husky **test** jobs from racing inline `bun run --filter=<dep> build` steps that used to wipe and rewrite shared upstream `dist/` trees, which caused flaky `"<name>" is not declared in this file"` failures.
> 
> **Hook behavior:** `phaseMaxConcurrency` centralizes per-phase limits (`test` → **1**; lint/typecheck stay parallel). Pre-commit and pre-push split typecheck vs test into two phases, skip tests when typecheck fails, and both route jobs through shared `runWithConcurrencyPool` so the cap is actually enforced (not just declared). Regression tests cover policy and pool overlap.
> 
> **Build output:** `diff`, `markdown`, `editor`, and `commentary` no longer `rm -rf dist` and write in place—they emit into unique `dist.tmp-*` dirs, validate outputs, then promote via `atomicSwapDist` (rename-based swap with concurrent-build handling). The helper is duplicated byte-for-byte in five packages (tsconfig `rootDir` constraint); `.gitignore` adds `dist.tmp-*` / `dist.old-*`. **Defense-in-depth** for CI/manual concurrent builds; hook serialization remains the primary gate fix.
> 
> **Tests:** Publication-contract concurrency fixture, mock FS race branches, duplication guard, and expanded `utilities.test.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3bf6cd93ae585d8b5f1e666ddeb39277ea9f8da2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,11 @@ jspm_packages/
 # Nuxt.js build output
 .nuxt/
 dist/
+# Staging / vacated build trees from atomicSwapDist (packages/*/scripts/build.ts).
+# Normally cleaned on exit; a hard kill (SIGKILL) can leave one behind, so ignore
+# them to keep `git status` clean and the commit/push flow unblocked.
+dist.tmp-*/
+dist.old-*/
 
 # SvelteKit build output
 .svelte-kit/
--- a/packages/commentary/scripts/build.ts
+++ b/packages/commentary/scripts/build.ts
@@ -1,20 +1,42 @@
 import { $ } from 'bun';
 
+import { atomicSwapDist } from './lib/atomic-swap-dist.ts';
 import { getExpectedBuildOutputs, readPackageExports } from './package-exports.js';
 
 const packageRoot = process.cwd();
+const distributionDirectory = `${packageRoot}/dist`;
+// Per-PID staging directory so concurrent same-package builds never collide on
+// a shared `dist.tmp` (each writer owns its own `dist.tmp-<pid>`).
+const stagingName = `dist.tmp-${process.pid}`;
+const stagingDirectory = `${packageRoot}/${stagingName}`;
 const packageExports = await readPackageExports();
 
-await $`rm -rf dist`;
+// Build to a staging directory first so a concurrent reader of `dist/` never
+// sees a partially-written tree. The `rm -rf dist` → write-to-dist pattern
+// opens a window where another process (e.g. a sibling package's test script
+// doing `bun run --filter=@cinder/commentary build`) wipes `dist` while the
+// first process's compiler is still writing into it. Building to
+// `dist.tmp-<pid>` then atomically renaming over `dist` closes that window: on
+// POSIX, `rename(2)` is atomic — a concurrent reader either sees the old tree
+// or the new tree, never a partial write. This is the root fix for issue #364.
+await $`rm -rf ${stagingDirectory}`;
 
-const compileResult = await $`tsc -p tsconfig.build.json`.nothrow();
+// `tsc` is configured to emit into `./dist` via tsconfig.build.json, but we
+// override `--outDir` on the command line so it emits into the staging
+// directory. The CLI flag takes precedence over the tsconfig value.
+const compileResult = await $`tsc -p tsconfig.build.json --outDir ./${stagingName}`.nothrow();
 if (compileResult.exitCode !== 0) {
   process.stderr.write(compileResult.stdout.toString());
   process.stderr.write(compileResult.stderr.toString());
   process.exit(1);
 }
 
-const expectedOutputs = getExpectedBuildOutputs(packageRoot, packageExports);
+// `getExpectedBuildOutputs` returns absolute paths rooted at the package root,
+// pointing into `dist/`. Substitute `dist` with the staging dir for the staging
+// validation — we must verify the staging tree before promoting it.
+const expectedOutputs = getExpectedBuildOutputs(packageRoot, packageExports).map((path) =>
+  path.replace(`${packageRoot}/dist/`, `${stagingDirectory}/`),
+);
 
 for (const outputPath of expectedOutputs) {
   if (!(await Bun.file(outputPath).exists())) {
@@ -22,5 +44,10 @@ for (const outputPath of expectedOutputs) {
     process.exit(1);
   }
 }
+
+// Atomically swap the staging directory into place without ever removing dist/
+// while it is live (see packages/diff/scripts/lib/atomic-swap-dist.ts for the
+// full concurrent-build rationale).
+atomicSwapDist(stagingDirectory, distributionDirectory);
 
 process.stdout.write('Build complete.\n');
--- a/packages/commentary/scripts/build.ts
+++ b/packages/commentary/scripts/build.ts
@@ -1,24 +1,25 @@
 import { $ } from 'bun';
 
-import { atomicSwapDist } from './lib/atomic-swap-dist.ts';
+import { atomicSwapDist, stagingDirectoryName } from './lib/atomic-swap-dist.ts';
 import { getExpectedBuildOutputs, readPackageExports } from './package-exports.js';
 
 const packageRoot = process.cwd();
 const distributionDirectory = `${packageRoot}/dist`;
-// Per-PID staging directory so concurrent same-package builds never collide on
-// a shared `dist.tmp` (each writer owns its own `dist.tmp-<pid>`).
-const stagingName = `dist.tmp-${process.pid}`;
+// Unique-per-invocation staging directory (a `dist.tmp-*` sibling of `dist`, so
+// the same filesystem `atomicSwapDist`'s rename requires). Each build owns its
+// own staging dir, so two concurrent same-package builds never collide on it.
+const stagingName = stagingDirectoryName();
 const stagingDirectory = `${packageRoot}/${stagingName}`;
 const packageExports = await readPackageExports();
 
-// Build to a staging directory first so a concurrent reader of `dist/` never
-// sees a partially-written tree. The `rm -rf dist` → write-to-dist pattern
-// opens a window where another process (e.g. a sibling package's test script
-// doing `bun run --filter=@cinder/commentary build`) wipes `dist` while the
-// first process's compiler is still writing into it. Building to
-// `dist.tmp-<pid>` then atomically renaming over `dist` closes that window: on
-// POSIX, `rename(2)` is atomic — a concurrent reader either sees the old tree
-// or the new tree, never a partial write. This is the root fix for issue #364.
+// Build into the staging directory, then promote it via `atomicSwapDist`. A
+// build never writes into `dist/` in place, so a concurrent reader (a sibling
+// package's test running `bun run --filter=@cinder/commentary build`, or a
+// typecheck holding dist declarations) can never observe a half-written tree.
+// NOTE: this is defense-in-depth, not the #364 fix — the husky test gate
+// serializes its test phase, and THAT is what stops concurrent same-package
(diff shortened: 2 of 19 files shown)


stevekinney:
Addressed all five review findings in `3bf6cd93`.

**Cursor Bugbot (Medium) — exit handler deletes rollback tree.** Real bug, confirmed via a cross-model second opinion. After step 2 vacates the live tree to `dist.old-*`, an *unexpected* step-3 install failure left `dist.old-*` as the sole copy of the last good build — and the `process.on('exit')` backstop then `rmSync`'d it, destroying the previous build. Two-part fix:
- Removed `rmSync(oldDirectory)` from the exit handler entirely. Path analysis shows it did useful work on **zero** paths: every safe path already removes `old` synchronously (step 4 success, the step-3 concurrent-loser path) or never creates it (step 1, step 2 ENOENT). Leaving a gitignored `dist.old-*` sibling behind on a hard kill is the safe trade.
- On the step-3 unexpected error, added a best-effort `rename(old → dist)` rollback before re-throwing the **original** error, so a failed install restores the previous build instead of leaving no `dist` at all. The rollback catch is empty-bodied: if it fails (e.g. a concurrent build installed a complete `dist` in the gap → `EEXIST`), we neither mask the original error nor delete `old`.
- New regression tests (`atomic-swap-dist.mock.test.ts`): rollback-success (asserts the 4th rename is `old→dist`), rollback-fail-preserves-old (original `ENOSPC` still thrown, `old` not deleted), and exit-handler-never-deletes-old (fires the registered listener, asserts it removes only `temp`).

**Copilot — canonical-source header said "four packages".** Corrected to describe the actual five-copy layout (components canonical + diff/markdown/editor/commentary), matching the duplication guard.

**Copilot — `pre-commit.ts` comment said "spurious TS1109 errors".** Corrected: #364's failure mode is a partial-tree read surfacing as `"<name>" is not declared in this file`, not the TS1109 parser error.

**Copilot — test spawned an external `mkdir`.** Added `mkdirSync` to the `node:fs` import and replaced `Bun.spawnSync(['mkdir', '-p', …])` with `mkdirSync(dir, { recursive: true })`.

All changes applied byte-identically across the five `atomic-swap-dist.ts` copies (duplication guard green). Local: atomic-swap suites 22 pass, utilities 106 pass, all five package typechecks 0 errors, prettier + oxlint clean.

PR #473, opened 2026-06-22
fix(components)!: require DataList key + narrow Masonry.as to layout-safe elements (#462)

### LINKED ISSUE #462: [audit][perf][HIGH] data-list has an unkeyed {#each} fallback when key is omitted
## Summary
`DataList` accepts an optional `key` prop and falls back to an unkeyed `{#each}` block when the prop is omitted. That makes a stable data-display primitive silently use index-based reconciliation for mutable lists. `Masonry` also types its `as` prop as any `string`, which allows void or otherwise inappropriate elements through `svelte:element` instead of constraining the public API to layout-safe element names.

## Why this is an issue
The unkeyed `DataList` path can create avoidable O(n) row churn and incorrect row instance reuse when consumers insert, remove, filter, or reorder items. Because the fallback is silent, consumers get no type-level signal that a stable key is required for correctness and performance. The broad `Masonry.as` type weakens developer experience and release quality by allowing element names that cannot validly contain masonry children.

## Scope
- Primary component scope: `packages/components/src/components/data-list/data-list.svelte`, `packages/components/src/components/data-list/data-list.types.ts`, and generated `data-list` schema and README artifacts.
- Related component scope: `packages/components/src/components/masonry/masonry.types.ts`, `packages/components/src/components/masonry/masonry.svelte` if needed, and generated `masonry` schema and README artifacts.
- Test scope: focused `data-list` and `masonry` tests or type tests that lock the new contracts.
- Out of scope: redesigning the `DataList` row snippet contract, adding a legacy compatibility shim for omitted keys, changing unrelated list primitives, or broadening this into a general layout-element audit.

## Acceptance criteria
- [ ] `DataListProps<T>` requires `key: (item: T) => string | number` instead of allowing `key` to be omitted.
- [ ] `data-list.svelte` renders non-empty `items` with a keyed `{#each items as entry (key(entry))}` block and contains no unkeyed `{#each items as entry}` fallback.
- [ ] Every repository-owned `DataList` usage, fixture, test, and example passes a stable key extractor.
- [ ] The generated `data-list` schema metadata marks `key` as a required unsupported function prop.
- [ ] A regression test or type test fails if `DataList` can be rendered without a `key` prop.
- [ ] `Masonry` exposes a named element union such as `MasonryElement`, and `MasonryProps.as` uses that union instead of `string`.
- [ ] The `MasonryElement` union excludes void elements such as `img`, `input`, `br`, and `hr`.
- [ ] A regression test or type test covers at least one allowed `Masonry.as` value and one disallowed void element value.

## Definition of done
- [ ] The unkeyed `DataList` branch is removed rather than hidden behind another fallback.
- [ ] The `Masonry.as` public type is narrowed without adding an adapter or compatibility layer.
- [ ] Focused tests or type tests cover the changed `DataList` and `Masonry` contracts.
- [ ] Generated component artifacts are updated after the source/type changes.
- [ ] The pull request notes the intentional breaking behavior for consumers that omitted `DataList.key`.
- [ ] No unrelated components, examples, or generated files are changed.

## Verification
```bash
bun run --filter=@lostgradient/cinder test -- ./src/components/data-list/data-list.test.ts ./src/components/masonry/masonry.test.ts
bun run --filter=@lostgradient/cinder components:generate
bun run --filter=@lostgradient/cinder components:check
bun run --filter=@lostgradient/cinder exports:check
bun run typecheck
bun run lint
```

## Original audit evidence
- Existing issue body reported this as **Lens: Architecture / performance** and **Severity: HIGH**.
- `packages/components/src/components/data-list/data-list.svelte` currently renders `items` with a keyed branch when `key` exists and an unkeyed fallback when it does not.
- `packages/components/src/components/data-list/data-list.types.ts` currently documents `key` as strongly recommended but optional.
- `packages/components/src/components/masonry/masonry.types.ts` currently declares `as?: string`.
- Original audit reference: `packages/components/docs/audits/component-audit-2026-06-20.md` sections B1 and B2.

stevekinney:
## Summary

Two breaking API-quality fixes:

1. **`DataList.key` is now required.** It was optional and silently fell back to an unkeyed `{#each}`, making a stable data-display primitive use index-based reconciliation for mutable lists — O(n) row churn and incorrect row instance reuse on insert/remove/filter/reorder. The unkeyed fallback is removed; `key` is required.
2. **`Masonry.as` is narrowed** from any `string` to a `MasonryElement` union (`article | aside | div | footer | header | main | nav | section | ul | ol`), excluding void elements (`img`/`input`/`br`/`hr`) and non-container elements that can't validly host masonry children.

Closes #462

## Breaking change

Both are intentional breaking changes (major bump, changeset included). Consumers that omitted `DataList.key` must pass a stable extractor (`key={(item) => item.id}`); consumers passing a void/inline element to `Masonry.as` get a compile error. No compatibility shims.

## Review committee

Pre-reviewed by: typescript-expert, svelte-expert, testing-expert + Codex (gpt-5.4, xhigh) — 1 round.
- **Must-fix addressed:** both regression tests were vacuous (a hardcoded local array / a positive-path render — they'd pass even if the fix were reverted). Replaced with real `@ts-expect-error` **type-level** guards: void+inline elements are not assignable to `MasonryElement`, and a props object omitting `DataList.key` is not assignable to `DataListProps`. Added a **reorder** test proving keyed reconciliation follows item identity (the old index fallback would fail it). typecheck passes with the directives correctly consumed (0 errors) — proving the contracts hold.
- svelte-expert verified all 10 repository `<DataList>` call sites pass a key; clean single-path breaking change, no hidden fallback.
- Codex APPROVED the implementation; flagged the vacuous tests (fixed) and noted `ul`/`ol` require `li` children (documented expectation).

## Test plan

```bash
bun run --filter=@lostgradient/cinder test -- ./src/components/data-list/data-list.test.ts ./src/components/masonry/masonry.test.ts
bun run --filter=@lostgradient/cinder components:check
bun run --filter=@lostgradient/cinder exports:check
bun run typecheck
```
20 pass / 0 fail; components:check + exports:check OK (regen drift-free after rebase on main); typecheck 0 errors.

---

> [!NOTE]
> **Medium Risk**
> Intentional major breaking changes: consumers omitting `DataList.key` or using invalid `Masonry.as` values must update or fail typecheck; list reconciliation behavior changes for anyone who relied on the old unkeyed fallback.
> 
> **Overview**
> **Major breaking API fixes** for `@lostgradient/cinder`: **`DataList` must receive `key`**, and **`Masonry.as` is limited to layout-safe container tags**.
> 
> **`DataList`:** `key` is required on `DataListProps` and always drives a keyed `{#each items as entry (key(entry))}`; the optional unkeyed `{#each}` path is removed so mutable lists cannot silently reconcile by index. Schemas, docs, examples, fixtures, and playground samples now pass a stable extractor (e.g. `key={(m) => m.id}`). Tests add a compile-time guard (`@ts-expect-error` when `key` is omitted) and a reorder behavior check.
> 
> **`Masonry`:** New `MasonryElement` union replaces `as?: string`, excluding void/inline tags like `img`, `input`, `br`, `hr`, `span`, and `button`. JSON schema uses an enum with default `"div"`. Tests assert allowed tags render and forbidden values fail type-check.
> 
> A **major** changeset documents both breaks; no runtime shims.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8713eb38564496e6e8e15f6e32bc5b027c526d9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


stevekinney:
Both addressed in the round-2 commit (f1e64e40):

- **data-list/README.md Usage snippet:** added `key={(member) => member.id}` — the hand-written usage block now passes the required `key`, so the copy-paste snippet compiles after the breaking change.
- **Masonry `as` default:** added `@default "div"` to the `as` JSDoc in `masonry.types.ts` (the generator source). This flows through to `masonry.schema.json` (`as.default: "div"`) and the README props table (`Default: "div"`), consistent with other components like Button.

Resolving.

stevekinney:
Addressed in a follow-up commit (a8713eb3). You're right that text-order alone doesn't distinguish keyed from unkeyed. The stronger node-identity assertion isn't reliable under happy-dom (it recreates nodes through Svelte's keyed-each reconciliation rather than moving them — verified: that assertion fails in this harness). So the comment now states this is an order/behavior check, and the load-bearing guard against 'key removed/made optional' is the compile-time `@ts-expect-error` test, which makes that regression a build error before it can reach runtime. Resolving.

stevekinney:
Both addressed: (1) reorder test — see the prior reply; the comment now states it's an order/behavior check and the load-bearing guard is the compile-time @ts-expect-error (node-identity isn't reliable under happy-dom). (2) stringKey now uses `unknown` instead of `any`+eslint-disable (a follow-up commit). Resolving.

PR #633, opened 2026-07-04
Allow Toolbar groups to wrap

### LINKED ISSUE #613: Toolbar overflows/collides at intermediate widths (wraps only at 30rem container breakpoint)
## Summary

`Toolbar` only switches from `flex-wrap: nowrap` to `wrap` at its own `@container cinder-toolbar (max-width: 30rem)` breakpoint. When a `Toolbar` is embedded in a wider flex container whose *available* width sits between 30rem and the toolbar's content width, the container query never fires, `min-inline-size: 0` lets the toolbar shrink below its content, and the buttons overflow and visually **overlap** instead of wrapping.

The 30rem threshold is measured against the toolbar's own inline size, not against whether its content actually fits — so any intermediate width where content > available but toolbar-width > 30rem is a dead zone that collides.

## Repro

```svelte
<div style="display:flex; gap:0.75rem; flex-wrap:wrap; width:720px;">
  <Toolbar aria-label="Actions">
    <Toolbar.Group>
      <Button size="sm" label="Place order" />
      <Button size="sm" label="Restaurant accepted" />
      <Button size="sm" label="Food ready" />
      <Button size="sm" label="Complete delivery" />
    </Toolbar.Group>
    <Toolbar.Group>
      <Button size="sm" label="Update address" />
      <Button size="sm" label="Get status" />
      <Button size="sm" label="List visibility" />
      <Button size="sm" label="Add tip" />
    </Toolbar.Group>
  </Toolbar>
</div>
```

At ~720px available width the two groups' content exceeds the box but the toolbar stays `nowrap` (its own width > 30rem), so buttons overlap. At >30rem-content the row is fine; at <30rem the container query rescues it. The gap in between collides.

## Expectation

A `Toolbar` embedded in a constrained container should degrade by wrapping its groups (as it already does below 30rem), not overflow/overlap. Options:

1. Make the horizontal toolbar `flex-wrap: wrap` by default (groups already stay `nowrap` internally until the 30rem query wraps their buttons too), or
2. Expose a `wrap` prop, or
3. Document that consumers must add `flex-wrap: wrap` themselves when embedding in a narrower-than-content container.

## Workaround in use

Overriding the component's layout from the consumer:

```css
.shell :global(.cinder-toolbar) {
  flex-wrap: wrap;
  row-gap: 0.5rem;
}
```

This makes groups wrap at intermediate widths and composes fine with the existing 30rem container query (which then wraps buttons within a group at the narrowest widths). Found while building the Sandman control strip (`@lostgradient/cinder@0.4.1`).

## Minor, related

`Toolbar.Group` renders `<div class="cinder-toolbar__group">` with no `role`. Since it's the semantic "grouping" primitive and accepts an `aria-label`, callers that want the group exposed to AT (and to `getByRole('group')` in tests) must pass `role="group"` themselves. Worth considering defaulting `role="group"` when an accessible name is present — or documenting that it's intentionally roleless.

stevekinney:
## Summary
- allow horizontal Toolbar roots to wrap groups by default so intermediate-width containers do not overlap controls
- keep individual Toolbar.Group contents nowrap until the existing narrow container query wraps group contents
- default named Toolbar.Group instances to role="group" and document the wrapping/group semantics

Fixes #613.

## Verification
- bun run --filter=@lostgradient/cinder test -- ./src/components/toolbar/toolbar.test.ts
- bun run --filter=@lostgradient/cinder components:check
- pre-commit hook
- pre-push scoped hook

---

> [!NOTE]
> **Medium Risk**
> Removes horizontal group dividers and changes wrap behavior, which can shift existing toolbar layouts; a11y role changes apply only to named groups.
> 
> **Overview**
> Horizontal **Toolbar** roots now use **`flex-wrap: wrap`** so **groups** move to new lines instead of overlapping in tight widths; each **`Toolbar.Group`** still **`nowrap`** until the existing **`@container cinder-toolbar (max-width: 30rem)`** rule wraps controls inside the group.
> 
> **Layout/visual:** pseudo-element **separators between adjacent horizontal groups are removed** (vertical group separators unchanged). **`Toolbar.Spacer`** + following group gets **`margin-inline-start: auto`** so trailing actions stay right-aligned when rows wrap.
> 
> **`Toolbar.Group`:** trims **`aria-label` / `aria-labelledby`**, drops empty name attrs, and sets **`role="group"`** when the group has an accessible name (or an explicit **`role`**). Docs, examples, and schema copy now describe wrapping instead of horizontal inter-group separators.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2769555b0342c68feff943d0e33b80a88e21a15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/components/src/components/toolbar/README.md
+++ b/packages/components/src/components/toolbar/README.md
@@ -22,6 +22,12 @@ A named grouping of related controls that implements the WAI-ARIA toolbar patter
 </Toolbar>
 ```
 
+Horizontal toolbars wrap groups by default so constrained containers do not
+overlap controls. Each `Toolbar.Group` keeps its controls together until the
+toolbar's narrow container query wraps controls inside the group. A named
+`Toolbar.Group` gets `role="group"` by default; pass an explicit `role` when a
+different semantic grouping is required.
+
 ## Props
 
 <!-- generated:props:start -->
--- a/packages/components/src/components/toolbar/README.md
+++ b/packages/components/src/components/toolbar/README.md
@@ -35,7 +35,7 @@ different semantic grouping is required.
 | Prop          | Type                           | Required | Default        | Description                                                                                                      |
 | ------------- | ------------------------------ | -------- | -------------- | ---------------------------------------------------------------------------------------------------------------- |
 | `class`       | `string`                       | no       | —              | Additional class merged with `.cinder-toolbar`.                                                                  |
-| `orientation` | `"horizontal"` \| `"vertical"` | no       | `"horizontal"` | Layout direction for keyboard ownership and separator placement.                                                 |
+| `orientation` | `"horizontal"` \| `"vertical"` | no       | `"horizontal"` | Layout direction for keyboard ownership and vertical group separator placement.                                  |
 | `children`    | `(opaque)`                     | yes      | —              | Controls rendered inside the toolbar. Not expressible in JSON Schema; see the component types for the signature. |
 
 <!-- generated:props:end -->
@@ -52,7 +52,7 @@ This component does not declare any local CSS variables.
 
 <!-- generated:subcomponents:start -->
 
-- `Toolbar.Group` — cluster related controls and let the toolbar draw separators between groups.
+- `Toolbar.Group` — cluster related controls and keep them together while the toolbar wraps.
 - `Toolbar.Spacer` — consume remaining space and push later groups to the far edge.
 
 <!-- generated:subcomponents:end -->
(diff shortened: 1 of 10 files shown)


stevekinney:
Review convergence summary:

- Kept horizontal toolbar wrapping as the default fix for #613 so intermediate-width containers wrap groups instead of overlapping controls.
- Removed the horizontal group-divider contract rather than moving dividers between flex items; CSS cannot reliably suppress separators only at flex-line breaks. Vertical group separators remain, and the Toolbar docs/examples now describe horizontal groups as wrapping clusters instead of separator owners.
- Normalized named Toolbar.Group ARIA attributes by trimming non-empty values and omitting empty strings.
- Broadened post-spacer group alignment from adjacent groups to all groups after the spacer so wrapped trailing groups remain end-aligned.
- Switched CSS selector assertions to whitespace-tolerant regex checks and added regression coverage for empty/trimmed Toolbar.Group accessible names.

PR #795, opened 2026-07-21
fix(ci): make required status checks reachable by every PR

stevekinney:
## Summary

Branch protection on `main` now requires four status checks: `unit-tests`, `typecheck`, `playwright`, and `Pre-1.0 changeset bump guard` (introduced by the in-flight `refactor/thin-local-gates` PR, which thins `pre-push` down to a fast fail-open hook and moves local validation entirely to CI + branch protection).

Three of the workflows backing those checks were path-filtered on `pull_request`. A required check with a path filter has a sharp edge: a PR whose changed files match none of the filtered paths never gets that check reported at all, and GitHub shows it stuck at "Expected" forever — permanently blocking merge. Concretely, any PR that doesn't touch `.changeset/**` (the common case — changesets are often added right before merge) would never trigger `changeset-guard.yaml`, so `Pre-1.0 changeset bump guard` would never run and the PR could never merge.

This landed as a live problem the moment branch protection went active — this PR fixes it directly so every in-flight and future PR can satisfy the required checks.

## Changes

- `unit-tests.yaml`: removed the `pull_request` path filter — runs on every PR unconditionally now.
- `changeset-guard.yaml`: removed the `pull_request` path filter for the same reason (it's cheap — install + one script — so unconditional is affordable).
- `browser-tests-skip.yaml` (new): mirror-skip workflow for `browser-tests.yaml`, which stays path-filtered (it's expensive — Chromium install + full Playwright run). Reports the identical `typecheck`/`playwright` job names on the inverse (`paths-ignore`) path set, succeeding immediately. For any diff, at least one of the two workflows reports each required name — sometimes both — never neither.

## Test plan

- [x] All three workflow YAML files parse (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] Verified against a live blocked PR (#794, the Changesets "Version Packages" PR) that required checks were missing entirely under the old path filters
- [ ] Confirm this PR's own checks (`unit-tests`, `typecheck`, `playwright`, `Pre-1.0 changeset bump guard`) all report and pass, proving the fix works end to end

Related: stevekinney/cinder refactor/thin-local-gates (adds the branch protection this fixes).

### DIFF
--- a/.github/workflows/browser-tests-skip.yaml
+++ b/.github/workflows/browser-tests-skip.yaml
@@ -1,0 +1,67 @@
+name: browser-tests-skip
+
+# Mirror-skip workflow for browser-tests.yaml's required status checks.
+#
+# `typecheck` and `playwright` (browser-tests.yaml) are required status
+# checks on `main` (see docs/validation-topology.md's "Branch protection"
+# section). browser-tests.yaml stays path-filtered — it's expensive
+# (Chromium install, a full Playwright run) — but a required check with a
+# path filter leaves any PR whose changes match NONE of that filter with the
+# check stuck at "Expected" forever, permanently blocking the merge.
+#
+# This workflow triggers on the exact INVERSE path set (`paths-ignore`
+# mirroring browser-tests.yaml's `paths`, entry for entry, negations
+# included) and reports the identical `typecheck`/`playwright` job names with
+# a job that succeeds immediately. For any given diff at least one of the two
+# workflows fires for each required name — sometimes both (accepted; the skip
+# job costs a few seconds of runner time), never neither.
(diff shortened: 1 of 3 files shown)


stevekinney:
Closing as fully superseded: the unfiltered changeset-guard trigger landed on main via the #798 line of work (with the stuck-required-check rationale documented in the workflow itself), unit-tests.yaml's trigger change was superseded by the #792/#798 turbo rewrite, and browser-tests-skip.yaml was the duplicate-check-name mirror pattern that #798 replaced with single-workflow job-level scope gating. Every reachability guarantee this PR targeted is now on main.

PR #844, opened 2026-07-22
fix(husky): remove stale pre-push utility residue

### LINKED ISSUE #836: husky utilities: dead helpers and stale gate-lock docs left behind by the pre-push thinning
#798 replaced the fat `pre-push` (scoped lint/typecheck/test dispatch behind a shared gate lock) with a fast fail-open sanity check. That was the right call, but it orphaned code and documentation in `packages/components/scripts/husky/utilities.ts` that still describes the old design. Review bots have flagged each piece independently across an unrelated PR, which is a good sign it is genuinely confusing rather than merely untidy.

## Dead exports

These have **no callers** anywhere outside `utilities.ts` and its own test file:

- `GateFailure`
- `summarizeFailures`
- `inferFailureScope`
- `formatFailureSummary`
- `writePrePushLog`

They existed to render the old hook's failure digests. Nothing renders a digest now.

```
$ git grep -l 'summarizeFailures' -- 'packages/**/*.ts' | grep -v 'utilities'
(no results)
```

## Comments describing a function that no longer exists

Several comments in `utilities.ts` and `utilities.test.ts` explain that the #364 concurrent-`dist` race is "closed at the source" by `preBuildDependencyClosure` in `pre-push.ts`. That function was deleted with the rest of the fat hook:

```
$ git grep -l 'preBuildDependencyClosure' -- 'packages/**/*.ts'
(no results)
```

So the comments now justify a mechanism that isn't there. Worth deciding explicitly whether the #364 protection still matters — the pre-build was the stated fix, and it is gone.

## "pre-push gate" naming for a now-generic lock

`withGateLock` / `withLocalValidationGateLock` is no longer used by `pre-push` at all. Its only remaining caller is `scripts/generate-component-artifacts.ts`. But:

- its doc comment still says `pre-push` holds the lock,
- its user-facing warnings and errors still say "pre-push gate", so someone hitting it while running `components:generate` gets a message naming a hook that isn't involved,
- `utilities.test.ts` asserts on that wording, so renaming means updating the assertion too.

## Related but separate

`parseWorkspaceOutputLine()` only recognizes `lint|typecheck|test` prefixes and never handled `test:changed:`, which `prePushPackageScript()` could produce. That predates #798 and is now largely moot since nothing dispatches those scripts from the hook — but it should be either fixed or deleted along with the rest of the digest machinery, not left half-alive.

## Why a separate issue

Surfaced while reviewing #799, which is a ~30-line detect-only change to the same file. Folding a five-export deletion plus a lock rename plus test updates into that PR would bury a small change under unrelated churn, and the decisions here (does #364 still need protection? what should the lock be called?) deserve their own consideration rather than being made in passing.

stevekinney:
## Summary

- remove dead pre-push failure digest helpers and their orphaned tests
- rename shared lock diagnostics for generic local validation callers
- preserve current live-holder lock behavior while removing stale dependency-closure documentation
- add a contract test that prevents the removed helpers from returning

## Validation

- `bun test packages/components/scripts/husky/utilities.test.ts packages/components/scripts/husky/pipeline-contract.test.ts` (123 passed)
- `bun run validate`

Closes #836


stevekinney:
@codex review

stevekinney:
@copilot review

stevekinney:
@codex review exact head 9d62842f1786a4c33ee95fd6c4f6fcc087c3a477

stevekinney:
@copilot review exact head 9d62842f1786a4c33ee95fd6c4f6fcc087c3a477

stevekinney:
@codex review exact head 4c08e28eb2e3c9e0a755a15c77142982e29e482d

stevekinney:
@copilot review exact head 4c08e28eb2e3c9e0a755a15c77142982e29e482d

PR #870, opened 2026-07-24
fix: expose StatusDot live labels as accessible names

### LINKED ISSUE #868: StatusDot live connection labels are not exposed as accessible names
## Problem

Tribunal issue stevekinney/tribunal#176 needs to replace a separate hidden stream-state label with Cinder `StatusDot` built-in labeling:

```svelte
<StatusDot connectionState={eventStreamConnectionState} label={connectionLabel} showLabel={false} size="sm" />
```

The Cinder schema says `label` is "used as the accessible name either way", and `connectionState` should keep live-region semantics. In both `@lostgradient/cinder@0.16.1` and `@lostgradient/cinder@0.17.0`, `StatusDot` suppresses `aria-label` whenever `resolvedLive` is true:

```svelte
role={resolvedLive ? 'status' : 'img'}
aria-live={resolvedLive ? 'polite' : undefined}
aria-atomic={resolvedLive ? 'true' : undefined}
aria-label={resolvedLive ? undefined : resolvedAriaLabel}
```

For live `role="status"`, the label is rendered as hidden child text, but the status region has no accessible name. A browser component assertion like this cannot find it:

```ts
await expect.element(page.getByRole('status', { name: 'streaming' })).toBeInTheDocument();
```

## Expected

`StatusDot` should preserve live-region behavior for `connectionState` and expose the provided `label`/`aria-label` as the accessible name when `showLabel={false}`.

## Actual

`StatusDot` preserves the live region but the live status region is unnamed. Downstream code must keep a separate hidden node or give up the accessible-name contract.

## Reproduction

1. In Tribunal with `@lostgradient/cinder@0.16.1`, render `StatusDot` with `connectionState`, `label`, and `showLabel={false}`.
2. Query with Playwright/Vitest Browser: `page.getByRole('status', { name: 'streaming' })`.
3. The query times out because the live status has no accessible name.

Verified the same source in the published `@lostgradient/cinder@0.17.0` tarball.

## Downstream blocker

Blocks stevekinney/tribunal#176.

stevekinney:
Closes #868

StatusDot keeps role=status and polite live-region semantics while exposing label or aria-label as the accessible name when showLabel={false}. Added focused rendering and role-query regressions.

### DIFF
--- a/.changeset/quiet-dots-name.md
+++ b/.changeset/quiet-dots-name.md
@@ -1,0 +1,5 @@
+---
+'@lostgradient/cinder': patch
+---
+
+Expose StatusDot labels as accessible names when live connection indicators hide their visible label.
--- a/packages/components/src/components/status-dot/status-dot.svelte
+++ b/packages/components/src/components/status-dot/status-dot.svelte
@@ -76,6 +76,13 @@
   const needsHiddenLiveLabel = $derived(
     resolvedLive && (!hasVisibleLabel || visibleLabelIsLiveDuplicate),
   );
+  // A live region with its visible label suppressed still needs an author name
+  // so role-based queries and assistive technology can identify the status.
+  const liveAriaLabel = $derived(
+    !hasVisibleLabel && (normalizedLabel !== undefined || normalizedAriaLabel !== undefined)
+      ? resolvedAriaLabel
+      : undefined,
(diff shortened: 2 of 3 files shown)


stevekinney:
@codex review

PR #902, opened 2026-07-24
fix(chat): invalidate stale push handlers after teardown

### LINKED ISSUE #899: Chat: push-handler closures stay live after unsubscribe — a misbehaving adapter can mutate typing/read-receipt state post-teardown
## Summary

Hardening request, low severity. When Chat tears down an adapter subscription (unmount or adapter/conversation swap), it calls the consumer's unsubscribe function but never invalidates the `ChatPushHandlers` closures it handed to `subscribe()`. An adapter that keeps a handler reference and calls it after its own unsubscribe (buffered event flushing late, a race in the consumer's transport teardown) still mutates Chat's internal typing/read-receipt/stream state for a subscription that no longer exists.

The documented contract implicitly requires adapters to self-gate; well-behaved adapters work fine (pinned by stevekinney/chatroom's `src/routes/exercises/adapter-push/subscription-lifecycle-fixture.svelte`, which clears its handler reference on teardown). But the failure mode is silent state corruption in exactly the class of async teardown race that's hardest for consumers to reason about.

## Requested fix

Invalidate the handler closures at unsubscribe time — have each handler check a per-subscription `active` flag and no-op (optionally `console.warn` in dev) once torn down. Zero cost to well-behaved adapters, converts a silent corruption into a no-op for everyone else.

stevekinney:
Closes #899\n\nInvalidate Chat push handlers during subscription teardown so late adapter callbacks cannot mutate state.

### DIFF
--- a/.changeset/899-stale-push-handlers.md
+++ b/.changeset/899-stale-push-handlers.md
@@ -1,0 +1,5 @@
+---
+'@lostgradient/chat': patch
+---
+
+Ignore push callbacks that arrive after a Chat adapter subscription is torn down.
--- a/packages/chat/src/lib/components/chat/container/chat.svelte
+++ b/packages/chat/src/lib/components/chat/container/chat.svelte
@@ -822,21 +822,32 @@
     if (!resolvedAdapter?.subscribe) return;
 
     const currentConversationId = conversationId;
+    let active = true;
     const handlers: ChatPushHandlers = {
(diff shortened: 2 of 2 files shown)


stevekinney:
@copilot Please review exact head b521c455a74e8828461ba5265482cfbad67c067d.

stevekinney:
Codex review request for exact head b521c455a74e8828461ba5265482cfbad67c067d: please re-check final implementation.

PR #975, opened 2026-07-25
fix(logo-cloud): honor requested columns

### LINKED ISSUE #924: LogoCloud: the columns prop is inert — values 3, 4, 5, and 6 all render two columns
## Symptom
Setting `columns={3}`, `{4}`, `{5}`, or `{6}` on `LogoCloud` renders a two-column grid regardless of the requested value.

## Root cause
`logo-cloud.css`:
- `.cinder-logo-cloud__list { … grid-template-columns: repeat(2, minmax(0,1fr)); }` (base rule)
- `[data-cinder-columns='3']`, `='4'`, `='5'`, `='6']` selectors all emit the **identical** `grid-template-columns: repeat(2, minmax(0,1fr))`.

These are copy-pasted variants where the repeat count was never updated to match the selector value.

## Fix
Correct each `[data-cinder-columns='N']` rule to emit `repeat(N, minmax(0,1fr))` matching its own selector.

## Acceptance criteria
- `columns={3}` computes `grid-template-columns: repeat(3, minmax(0, 1fr))` (and so on for 4, 5, 6).
- `columns={2}` (and default) remains `repeat(2, minmax(0, 1fr))`.

## Regression test
Add a test that renders LogoCloud with each supported `columns` value and asserts the computed `grid-template-columns` differs per value (not just that the attribute is set).

## Structural follow-up
LogoCloud hand-rolls its own grid instead of composing `Grid`, whose numeric `columns` prop resolves to `repeat(<n>, …)` and cannot desynchronize this way. Tracked under #919.

stevekinney:
## Summary

- correct LogoCloud base grid selectors for column counts 3 through 6
- add a real-browser computed-style regression covering the default and every supported count
- add a patch changeset

## Verification

- `bun test ./packages/components/src/components/logo-cloud/logo-cloud.test.ts`
- `./node_modules/.bin/playwright test --config=packages/testing/playwright.config.ts packages/testing/tests/logo-cloud-columns.playwright.ts`
- `bun run --filter=@cinder/testing typecheck`
- `bunx stylelint packages/components/src/components/logo-cloud/logo-cloud.css`
- `bunx oxlint packages/testing/tests/logo-cloud-columns.playwright.ts`

Closes #924

### DIFF
--- a/.changeset/fuzzy-logos-count.md
+++ b/.changeset/fuzzy-logos-count.md
@@ -1,0 +1,5 @@
+---
+'@lostgradient/cinder': patch
+---
+
+Make LogoCloud honor column counts from three through six at its base container size.
--- a/packages/components/src/components/logo-cloud/logo-cloud.css
+++ b/packages/components/src/components/logo-cloud/logo-cloud.css
@@ -36,19 +36,19 @@
   }
 
   .cinder-logo-cloud[data-cinder-columns='3'] .cinder-logo-cloud__list {
-    grid-template-columns: repeat(2, minmax(0, 1fr));
+    grid-template-columns: repeat(3, minmax(0, 1fr));
   }
 
   .cinder-logo-cloud[data-cinder-columns='4'] .cinder-logo-cloud__list {
-    grid-template-columns: repeat(2, minmax(0, 1fr));
+    grid-template-columns: repeat(4, minmax(0, 1fr));
   }
 
   .cinder-logo-cloud[data-cinder-columns='5'] .cinder-logo-cloud__list {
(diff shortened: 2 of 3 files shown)


stevekinney:
Closing after source-of-truth validation. `LogoCloudProps.columns` and the generated schema define this as the grid count **on wide screens**. The two-column base is the intended narrow-container cap; the existing container queries increase to 3/4 at 48rem and 5/6 at 64rem. Changing the base selectors made the default five-column layout cramped on mobile and caused widening from below 48rem to reduce five/six tracks back to four. The audit finding misclassified responsive behavior as an inert prop, so this change must not merge.

PR #1037, opened 2026-07-27
Make the downstream snapshot safe for unattended audits

### LINKED ISSUE #901: Add a deterministic downstream-consumer snapshot for Cinder audits
## Observed repeated work

The daily downstream-usage automation performed the same mechanical evidence collection on each of eight runs from July 17 through July 24:

- resolve the published `@lostgradient/cinder` and `@lostgradient/chat` versions;
- inspect the published tarballs;
- resolve the current remote commit for each downstream repository;
- scan manifests and lockfiles for declared and locked versions;
- scan imports, style entrypoints, private selectors, and Cinder workarounds;
- collect open and recently closed issues before semantic deduplication.

The active automation now covers six downstream repositories. Four paused copies of the same automation retain an older five-repository prompt. The repeated extraction is currently performed by models and returned as prose rather than by a canonical, mechanically verifiable Cinder tool.

This has already had a reliability cost: the July 21 downstream audit created an Agent Bureau issue and then closed it after discovering the canonical issue had been reopened.

No existing repository script, project skill, or workflow provides a downstream snapshot or issue-deduplication input for these audits. Existing root scripts are limited to the next-release, preload, SSR-boundary, and pull-request-triage workflows.

## Impact

Every run spends tool calls and context on unchanged repository discovery and raw file reads before the model can do the useful semantic work. Prose-shaped evidence is difficult to diff between runs, makes stale-ref mistakes harder to detect, and increases the chance of duplicate issue creation.

Exact token savings cannot be calculated from current automation metadata because per-step token attribution is unavailable. The measured frequency is eight executions in the audit window, across six repositories plus two published packages.

## Root cause

**Confidence: high.** Stable extraction and normalization work is embedded in the automation prompt instead of being owned by a deterministic repository tool. The model is therefore re-discovering and reshaping the same facts on every run.

## Bounded implementation scope

Add `scripts/cinder-downstream-snapshot.ts` and `scripts/cinder-downstream-snapshot.test.ts`.

The tool must remain read-only and must not classify findings or write GitHub issues. Those semantic and authorization-sensitive decisions stay with the coordinator.

### Input

Accept a versioned JSON request containing:

- npm package names and either an explicit version or a resolution policy;
- downstream repository names and explicit commit refs, or a branch whose resolved commit is recorded;
- include/exclude globs for manifests, lockfiles, source imports, style entrypoints, private selectors, and known workaround markers;
- the bounded open/recently-closed issue window.

### Output

Write sorted, compact JSON to stdout with:

- schema version and collection timestamp;
- each resolved npm version, tarball integrity, and relevant exports;
- each requested repository ref and resolved commit SHA;
- declared and locked Cinder package versions;
- matching imports, styles, selectors, workarounds, and issue-candidate evidence with exact paths and line numbers;
- collection errors separated from an otherwise successful snapshot.

Do not emit model-written conclusions. A second run against the same package versions and resolved commit SHAs must be byte-identical apart from an explicitly excluded collection timestamp.

## Acceptance criteria

- [ ] The script performs no repository or GitHub writes.
- [ ] Tests cover npm metadata/tarball extraction, mutable-branch resolution, manifest and lockfile parsing, source/style evidence, issue-window normalization, partial collection failures, stable sorting, and JSON schema versioning.
- [ ] Replaying the same resolved package versions and commit SHAs produces identical normalized evidence.
- [ ] The output is compact enough for an automation to consume without rereading the underlying files.
- [ ] Semantic equivalence, root-cause analysis, prioritization, and issue creation remain outside the script.
- [ ] The daily and weekly Cinder audits can invoke the same script rather than maintaining separate extraction prompts.

## Verification

```bash
bun test scripts/cinder-downstream-snapshot.test.ts
bun run scripts/cinder-downstream-snapshot.ts --help
bun run scripts/cinder-downstream-snapshot.ts --request ./scripts/fixtures/cinder-downstream-snapshot/request.json --output /tmp/cinder-downstream-snapshot.json
bun run scripts/cinder-downstream-snapshot.ts --request ./scripts/fixtures/cinder-downstream-snapshot/request.json --output /tmp/cinder-downstream-snapshot-second.json
diff -u /tmp/cinder-downstream-snapshot.json /tmp/cinder-downstream-snapshot-second.json
```

---

## Reopened: the snapshot is not yet safe for unattended automation

The shipped collector accepts only local repository paths. A scheduled audit must
analyze each repository's **current default branch**, not whichever branch happens
to be checked out locally.

At the time this requirement was discovered, `/Users/stevekinney/Developer/tribunal`
was on a topic branch rather than `main`. Scanning that checkout would produce
stale or branch-specific evidence while claiming to audit current downstream usage.
The existing `branch` field does not help: it only *asserts* that the local
checkout already sits on the requested branch, and errors out otherwise. It cannot
fetch the current default-branch state.

There is also no canonical production request. The only committed request is
`scripts/fixtures/cinder-downstream-snapshot/request.json`, which is a test
fixture pointed at this repository.

The collector therefore needs a deterministic remote/ref source mode and a
canonical production request before any scheduled automation can safely adopt it.

## Additional acceptance criteria

- [ ] Each repository request accepts exactly one of a local checkout
      (`path`, optional `branch`) or a remote source (`remote` + `ref`).
- [ ] The request is rejected — non-zero exit, offenders named — when an entry
      supplies both `path` and `remote`, neither, `remote` without `ref`, or
      `ref` without `remote`.
- [ ] A remote request clones only the requested ref into a uniquely named
      temporary directory, prefers shallow single-branch cloning, resolves and
      records the exact commit SHA, and runs the existing collector against that
      temporary checkout.
- [ ] The temporary checkout is removed in a `finally` block after both success
      and failure. No existing source repository is ever written to.
- [ ] Output includes `remote`, `ref`, and the resolved `commit`; `.git` stays
      excluded; stable sorting, partial-error reporting, hashing, and secret
      redaction are retained.
- [ ] Clone failures are not hidden: they appear in `errors` with scope
      `repository:<name>` and the run still exits zero.
- [ ] `scripts/cinder-downstream-audit-request.json` exists as the canonical
      production request covering the six downstream repositories and the two
      published packages, with each `ref` matching the repository's live default
      branch.
- [ ] Tests use temporary local Git repositories or `file://` remotes. They do
      not call GitHub or npm.

stevekinney:
## Summary

`scripts/cinder-downstream-snapshot.ts` only accepted local repository paths, so a
scheduled audit analyzed whichever branch happened to be checked out locally rather
than each downstream repository's current default branch. When this was discovered,
`~/Developer/tribunal` sat on a topic branch — scanning it would have produced
stale, branch-specific evidence while claiming to audit current downstream usage.
The existing `branch` field could not help: it only *asserts* that a checkout
already sits on the requested branch and errors out otherwise.

There was also no canonical production request. The only committed request is
`scripts/fixtures/cinder-downstream-snapshot/request.json`, a test fixture pointed
at this repository.

## What changed

A repository request now declares exactly **one** source:

```json
{ "name": "repository", "path": "/absolute/path", "branch": "main" }
{ "name": "repository", "remote": "https://github.com/owner/repository.git", "ref": "main" }
```

Malformed entries are rejected at the boundary — before any collection work — with a
nonzero exit that names every offender: both `path` and `remote`, neither, `remote`
without `ref`, and `ref` without `remote`.

For a remote request the collector shallow-clones only the requested ref
(`--depth 1 --single-branch --branch <ref>`) into a uniquely named temporary
directory, resolves the exact commit SHA, runs the existing evidence collector
against that clone, and removes the checkout in a `finally` block. No existing
source repository is ever written to, and no locally checked-out branch can
influence the result. Output gains `remote`, `ref`, and the resolved `commit`;
`.git` exclusion, stable sorting, hashing, partial-error reporting, and secret
redaction are unchanged.

Clone failures are not hidden — they land in `errors` with scope
`repository:<name>` and the run still exits zero, so one unreachable repository
cannot void an otherwise complete snapshot.

A remote request is also rejected — not silently narrowed — when it carries
`branch` or `commit`, since those pin a local checkout only and accepting them
would let a misconfigured audit believe it pinned a revision it did not. HTTP(S)
remotes with embedded userinfo (`https://user:token@…`) are rejected too: the
remote string is written verbatim into `repositories[].remote` and into any
clone-failure message, so accepting one would put a token in the artifact these
audits are meant to keep and diff. SSH remotes carry a user name rather than a
secret and stay valid, as do `file://` remotes.

Clone cleanup is best-effort. A throw from `rm()` inside `finally` would escape
the per-repository `catch` that already handled the real failure, aborting the
whole run and emitting no snapshot at all — so an unremovable checkout is
reported as a `repository:<name>` error instead, and every other repository
still collects.

`scripts/cinder-downstream-audit-request.json` is the new canonical production
request: the two published packages plus the six downstream repositories, each
`ref` verified against the repository's live GitHub default branch.

Refactoring to support this extracted `collectRepositoryEvidence`,
`resolveLocalCommit`, `resolveHeadCommit`, `cloneRemoteRef`, and `collectPackages`
out of the previously monolithic `main`, and typed the snapshot accumulators that
TypeScript had been inferring as `never[]`. Local `path`/`branch`/`commit`
semantics are unchanged.

## Tests

Twenty-two tests, all built on the `runSnapshotCli` helper from #1033. New coverage:
valid local input, valid remote/ref input, all four invalid source combinations,
exact commit recording, temporary-clone cleanup after success and after failure,
`.git` exclusion, stable ordering across remotes, clone failure landing in `errors`,
no mutation of the source repository (HEAD, status, index, and reflog compared
before and after), byte-identical replay of a remote request against the same
commit, rejection of remote `branch`/`commit` pins and credential-bearing
remotes, best-effort cleanup reporting an unremovable checkout instead of
throwing, and the canonical request's shape. Tests use temporary local Git
repositories and `file://` remotes — no GitHub or npm calls. Cleanup assertions
compare the temporary-directory listing before and after a run rather than an
absolute count, so concurrent worktrees cannot flake them.

## Validation

```
bun run test:workspace-scripts        22 pass, 0 fail
bunx oxlint --config .oxlintrc.json   0 warnings, 0 errors
prettier --check                      clean
tsc --noEmit --strict                 clean
git diff --check origin/main..HEAD    clean
```

The canonical request ran end to end against all six live repositories: zero
errors, six repositories, every commit resolved, and no temporary clone directories
left behind.

| repository | ref | resolved commit |
| --- | --- | --- |
| agent-bureau | main | `d9da85eb37fe6854a272841949210e2ded2b70a6` |
| sandman | main | `a2b105e354f50ebaa29f36437b9a62558f507ae9` |
| stardust | main | `9a1402f9e2bc0098d533d45557f9af2b97e5ce33` |
| temporal-explorer | main | `ad1200e1f42bdfa86a14bc977799d7223c6cb606` |
| tribunal | main | `1db649e6bcc27ad92ac689e6e41128b4f4064bda` |
| weft-console | main | `dd94ae88f50bd310ce04b411932769cf172f406d` |

Resolved packages: `@lostgradient/chat@0.4.1`, `@lostgradient/cinder@0.19.1`.
The tribunal commit is `main`, not the topic branch its local checkout is on —
which is the whole point of the change.

Closes #901

### DIFF
--- a/scripts/cinder-downstream-audit-request.json
+++ b/scripts/cinder-downstream-audit-request.json
@@ -1,0 +1,165 @@
+{
+  "schemaVersion": 1,
+  "packages": [
+    {
+      "name": "@lostgradient/chat",
+      "resolution": "published"
+    },
+    {
+      "name": "@lostgradient/cinder",
+      "resolution": "published"
+    }
+  ],
+  "repositories": [
+    {
+      "name": "agent-bureau",
+      "remote": "https://github.com/stevekinney/agent-bureau.git",
+      "ref": "main",
+      "globs": [
+        "**/package.json",
+        "**/bun.lock",
+        "**/pnpm-lock.yaml",
+        "**/package-lock.json",
+        "**/yarn.lock",
+        "**/*.{ts,tsx,js,jsx,svelte,css,scss,md,mdx}"
+      ],
+      "evidence": {
+        "manifests": [
+          "**/package.json",
+          "**/bun.lock",
+          "**/pnpm-lock.yaml",
+          "**/package-lock.json",
+          "**/yarn.lock"
+        ],
+        "source": ["**/*.{ts,tsx,js,jsx,svelte}"],
+        "styles": ["**/*.{css,scss}"],
+        "documentation": ["**/*.{md,mdx}"]
+      }
+    },
+    {
+      "name": "sandman",
+      "remote": "https://github.com/stevekinney/sandman.git",
+      "ref": "main",
+      "globs": [
+        "**/package.json",
+        "**/bun.lock",
+        "**/pnpm-lock.yaml",
+        "**/package-lock.json",
+        "**/yarn.lock",
+        "**/*.{ts,tsx,js,jsx,svelte,css,scss,md,mdx}"
+      ],
+      "evidence": {
+        "manifests": [
+          "**/package.json",
+          "**/bun.lock",
+          "**/pnpm-lock.yaml",
+          "**/package-lock.json",
+          "**/yarn.lock"
+        ],
+        "source": ["**/*.{ts,tsx,js,jsx,svelte}"],
+        "styles": ["**/*.{css,scss}"],
+        "documentation": ["**/*.{md,mdx}"]
+      }
+    },
+    {
+      "name": "stardust",
+      "remote": "https://github.com/stevekinney/stardust.git",
+      "ref": "main",
+      "globs": [
+        "**/package.json",
+        "**/bun.lock",
+        "**/pnpm-lock.yaml",
+        "**/package-lock.json",
+        "**/yarn.lock",
+        "**/*.{ts,tsx,js,jsx,svelte,css,scss,md,mdx}"
+      ],
+      "evidence": {
+        "manifests": [
+          "**/package.json",
+          "**/bun.lock",
+          "**/pnpm-lock.yaml",
+          "**/package-lock.json",
+          "**/yarn.lock"
+        ],
+        "source": ["**/*.{ts,tsx,js,jsx,svelte}"],
+        "styles": ["**/*.{css,scss}"],
+        "documentation": ["**/*.{md,mdx}"]
+      }
+    },
+    {
+      "name": "temporal-explorer",
+      "remote": "https://github.com/stevekinney/temporal-explorer.git",
+      "ref": "main",
+      "globs": [
+        "**/package.json",
+        "**/bun.lock",
+        "**/pnpm-lock.yaml",
+        "**/package-lock.json",
+        "**/yarn.lock",
+        "**/*.{ts,tsx,js,jsx,svelte,css,scss,md,mdx}"
+      ],
+      "evidence": {
+        "manifests": [
+          "**/package.json",
+          "**/bun.lock",
+          "**/pnpm-lock.yaml",
+          "**/package-lock.json",
+          "**/yarn.lock"
+        ],
+        "source": ["**/*.{ts,tsx,js,jsx,svelte}"],
+        "styles": ["**/*.{css,scss}"],
+        "documentation": ["**/*.{md,mdx}"]
+      }
+    },
+    {
+      "name": "tribunal",
+      "remote": "https://github.com/stevekinney/tribunal.git",
+      "ref": "main",
+      "globs": [
+        "**/package.json",
+        "**/bun.lock",
+        "**/pnpm-lock.yaml",
+        "**/package-lock.json",
+        "**/yarn.lock",
+        "**/*.{ts,tsx,js,jsx,svelte,css,scss,md,mdx}"
+      ],
+      "evidence": {
+        "manifests": [
+          "**/package.json",
+          "**/bun.lock",
+          "**/pnpm-lock.yaml",
+          "**/package-lock.json",
+          "**/yarn.lock"
+        ],
+        "source": ["**/*.{ts,tsx,js,jsx,svelte}"],
+        "styles": ["**/*.{css,scss}"],
+        "documentation": ["**/*.{md,mdx}"]
+      }
+    },
+    {
+      "name": "weft-console",
+      "remote": "https://github.com/stevekinney/weft-console.git",
+      "ref": "main",
+      "globs": [
+        "**/package.json",
+        "**/bun.lock",
+        "**/pnpm-lock.yaml",
+        "**/package-lock.json",
+        "**/yarn.lock",
+        "**/*.{ts,tsx,js,jsx,svelte,css,scss,md,mdx}"
+      ],
+      "evidence": {
+        "manifests": [
+          "**/package.json",
+          "**/bun.lock",
+          "**/pnpm-lock.yaml",
+          "**/package-lock.json",
+          "**/yarn.lock"
+        ],
+        "source": ["**/*.{ts,tsx,js,jsx,svelte}"],
+        "styles": ["**/*.{css,scss}"],
+        "documentation": ["**/*.{md,mdx}"]
+      }
+    }
+  ]
+}
--- a/scripts/cinder-downstream-snapshot.test.ts
+++ b/scripts/cinder-downstream-snapshot.test.ts
@@ -1,10 +1,15 @@
 import { expect, test } from 'bun:test';
-import { mkdir, mkdtemp } from 'node:fs/promises';
+import { chmod, mkdir, mkdtemp, readdir } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
-import { selectMostRecentlyPublishedVersion } from './cinder-downstream-snapshot.ts';
+import {
+  CLONE_DIRECTORY_PREFIX,
+  describeInvalidRepositorySource,
+  removeTemporaryCheckout,
+  selectMostRecentlyPublishedVersion,
+} from './cinder-downstream-snapshot.ts';
 
 /**
  * Absolute path to the CLI under test. Resolving from `import.meta.url` rather than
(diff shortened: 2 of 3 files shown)


stevekinney:
Good catch — fixed in \`8ccc9d13e\`.

A throw from `rm()` inside `finally` does not just override the caught error; it escapes the per-repository `catch` entirely (that handler has already run), so a single unremovable checkout would abort the whole run and emit no snapshot JSON at all. That is precisely the unattended-automation failure this change exists to prevent.

Cleanup is now best-effort via `removeTemporaryCheckout()`, which returns a `CollectionError` rather than throwing. The caller pushes it onto `errors` with scope `repository:<name>`, so an unremovable checkout is reported, the remaining repositories still collect, the snapshot is still written, and the exit stays zero — consistent with how clone failures are already handled.

Covered by a new test that removes a checkout successfully, then `chmod 0o500` on the parent so the child entry cannot be unlinked, and asserts the returned error scope and message instead of a throw. 21 tests pass.

stevekinney:
Both fixed in `e202efdda`.

**Credential-bearing remotes and silently-ignored pins.** You are right that this was a leak, not just untidiness — the remote string is echoed verbatim into `repositories[].remote` *and* into the clone-failure message, so an `https://user:token@…` remote would write a token into the very artifact these audits are meant to keep and diff. The existing redaction only covers secret-shaped assignments inside scanned evidence lines; it never saw the request. `describeInvalidRepositorySource` now rejects a remote request that carries `branch` or `commit` (those pin a local checkout only, so accepting them lets a misconfigured audit believe it pinned a revision it did not) and rejects an HTTP(S) remote with userinfo, pointing at a credential helper instead. The check is scoped to `http:`/`https:` on purpose: `git@github.com:owner/repository.git` carries a user name rather than a secret, and `file://` remotes carry no userinfo, so both stay valid. Tests cover `user:token@`, bare `token@`, plain `http://`, and both non-rejection cases.

**Duplicated `CLONE_DIRECTORY_PREFIX`.** Now exported from `cinder-downstream-snapshot.ts` and imported by the tests, so the leak-detection assertions cannot drift into false negatives.

Also extended the canonical-request guard to assert no entry carries `branch` or `commit`. 22 tests pass; oxlint, prettier, and `tsc --strict` clean; the canonical request still runs against all six live remotes with zero errors, every commit resolved, byte-identical to earlier runs apart from `collectedAt`, and no clone directories left behind.

stevekinney:
Resolving the four `toSorted()` threads without changing the code — the premise does not hold for this file, and I checked rather than assuming.

The claim is that the repo targets ES2022 and therefore avoids `toSorted()`. The root `tsconfig.json` sets `"lib": ["ESNext", "DOM", "DOM.Iterable"]` alongside `"target": "ES2022"`, and `lib` — not `target` — is what governs which built-ins are available. `target` only controls syntax downleveling, and with `"noEmit": true` nothing is downleveled at all. `tsc --strict` on this file is clean.

The convention claim does not survive contact with the codebase either:

- `toSorted()` has 130 call sites across `packages/` and `scripts/`.
- 13 of them are in `packages/components/src` — the published package the cited comment says cannot use it.
- `packages/markdown/src/diff/compute.ts:237` is also published source.
- `scripts/cinder-downstream-snapshot.ts` already had **six** `toSorted()` calls on `main` before this PR. If the hazard were real for this file, `main` would already carry it.

The cited precedent in `phone-input-formatting.ts` is a decision local to that one file, not a repo-wide rule. These scripts run only under Bun, locally and in CI, and are never shipped to a consumer runtime — `toSorted()` has been available in Bun since 1.0.

Rewriting six pre-existing and several new call sites to `Array.from(...).sort()` late in a review cycle, on a premise 130 call sites contradict, would add churn without removing a defect. Leaving them.

Separately and out of scope for #901: that `phone-input-formatting.ts` comment asserts a constraint the rest of the repo — including 13 call sites in the published component source — does not observe. Either the comment is stale or those call sites are a genuine shipping hazard. Worth resolving deliberately in its own issue rather than absorbed here.

PR #1065, opened 2026-07-31
fix(speed-dial): bridge portaled Tab boundaries

### LINKED ISSUE #1050: SpeedDial Tab bridging is heuristic, not a real sequential-focus model
## Context

PR #976 fixed the most direct SpeedDial Tab-boundary bug (a `tabindex="-1"`
first/last action being treated as the Tab boundary). Follow-up review
(chatgpt-codex-connector) on that same PR found two more gaps in the same
area that are not fixed there, because closing them properly means building
a real sequential-focus model for the portaled toolbar rather than patching
another special case onto the existing heuristics:

1. **Arrow-focused untabbable edge actions still escape wrong** (flagged on
   `speed-dial.svelte:270`): if the first action has `tabindex="-1"` and is
   reached via arrow keys (not native Tab), Shift+Tab from it does not match
   `getSequentiallyTabbableActionButtons()`'s boundary check (that action
   isn't in the sequentially-tabbable list at all), so the bridge doesn't
   fire and native Tab resolves from the toolbar's portaled DOM position
   instead of before the SpeedDial. Symmetric gap for a last untabbable
   action on forward Tab.

2. **Reverse Tab from the open trigger doesn't return to the last action**
   (flagged on `speed-dial.svelte:88`): `handleTriggerKeydown` only handles
   arrow keys, never Shift+Tab. Because the actions toolbar is portaled out
   of the trigger's native DOM position, Shift+Tab from the trigger while
   open jumps to whatever precedes the SpeedDial on the page instead of the
   last open action — the reverse of the natural "toolbar before trigger"
   order in markup.

## Relationship to #1048

https://github.com/stevekinney/cinder/issues/1048 tracks the same
underlying problem for NavigationBar and SpeedDial's *forward-adjacency*
whitelist (missing native elements like `<summary>`/`<iframe>`). This issue
is the SpeedDial-specific *internal* Tab-sequence gaps (arrow-focus vs.
Tab-boundary state, and trigger-to-toolbar reverse bridging) that remain
even with a complete whitelist. Both point at the same fix: replace the
ad hoc boundary/whitelist checks with an actual sequential-focus model for
portaled interactive regions.

## Proposed fix

Track "is the currently focused action before/at/after the sequentially
tabbable boundary" as real state (not just `target === tabOrderButtons[0]`),
and add a `Shift+Tab` handler on the trigger while `open` that intercepts
and moves focus to the last sequentially tabbable action.

stevekinney:
## Summary

- bridge Tab and Shift+Tab across the portaled SpeedDial toolbar, including arrow-focused `tabindex="-1"` actions
- leave native navigation intact when every action is sequentially untabbable or the portaled toolbar is inert
- derive boundaries from current toolbar DOM order and avoid repeated rendered-candidate scans
- document the focus keyboard matrix and add regression coverage

Closes #1050

## Validation

Exact rebased head: `60c9ff9f0fdb51d0ac854438d97056e394f71513`

- `TZ=UTC LANG=en_US.UTF-8 bun test --preload ./scripts/preload.ts --conditions browser --conditions svelte --parallel=1 src/components/speed-dial/speed-dial.test.ts src/_internal/anchored-overlay.test.ts` - 46 pass
- `bun run --filter=@lostgradient/cinder components:check`
- `bun run --filter=@lostgradient/cinder typecheck` - 0 errors
- `bun run --filter=@lostgradient/cinder lint:invariants`
- targeted Prettier and `git diff --check`
- pre-push fast sanity checks

The earlier Playwright failure was on the pre-mainline branch and exercised SelectionPopover, not SpeedDial. This exact head is rebased onto current `origin/main`, which contains the merged SelectionPopover and anchored-overlay fixes; fresh exact-head CI is the merge gate.

### DIFF
--- a/.changeset/portaled-speed-dial-tab-bridge.md
+++ b/.changeset/portaled-speed-dial-tab-bridge.md
@@ -1,0 +1,5 @@
+---
+'@lostgradient/cinder': patch
+---
+
+Keep SpeedDial Tab navigation contiguous across its portaled actions, including arrow-focused untabbable actions.
--- a/packages/components/src/components/speed-dial/speed-dial.a11y.md
+++ b/packages/components/src/components/speed-dial/speed-dial.a11y.md
@@ -10,7 +10,23 @@ Keyboard behavior:
 - Enter or Space on the trigger toggles the dial through the native button.
 - Arrow keys on the trigger open the dial and focus the first enabled action.
 - Arrow keys, Home, and End move through enabled actions in the toolbar.
+- Tab from the final sequentially tabbable action returns focus to the trigger;
+  the same bridge applies when the focused final action is `tabindex="-1"` and
+  was reached with arrow-key navigation.
+- Shift+Tab from the first sequentially tabbable action, or from an earlier
+  arrow-focused action, returns focus to the nearest sequentially focusable
+  element before the SpeedDial. Shift+Tab from the open trigger focuses the
+  last sequentially tabbable action.
 - Escape closes the dial and returns focus to the trigger.
 
 Closed actions are kept mounted but inert so exit motion can complete without
 leaving focusable descendants behind.
+
+## Focus interaction review
+
+The portaled toolbar is treated as one sequential-focus region even though its
+DOM position is outside the trigger. The keyboard matrix above covers native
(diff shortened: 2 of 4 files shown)


stevekinney:
@codex review

PR #1066, opened 2026-07-31
fix(carousel): settle cancelled native gestures after release

### LINKED ISSUE #1053: carousel: isInteracting/isNativeScrolling pointer-state edge cases (native-scroll follow-ups)
## Context

PR #1010 (native touch scrolling and scroll snapping for Carousel, closing #957) went through an extended automated review cycle. Six rounds of `chatgpt-codex-connector` findings on the pointer/scroll interaction state machine were fixed with code changes and regression tests in that PR:

- Vertical wheel events no longer cancel a pending programmatic/autoplay scroll (`onWheel`).
- Reordering the `slides` array while resting re-anchors `settledIndex` from the DOM instead of trusting a stale numeric index.
- An unrelated window blur (no tracked pointer interaction) no longer cancels a pending programmatic destination.
- Wheel-cancels-scroll now requires horizontal dominance (`|deltaX| > |deltaY|`), not just nonzero `deltaX`.
- Only `touch`/`pen` pointers start a trackable interaction (`onPointerDown`), not `mouse`.
- The interaction-layout window (which keeps physical neighbor slides laid out) only widens once the track is actually moving (`isNativeScrolling` or `programmaticTarget`), not merely because a pointer is held down.

A seventh finding surfaced after that sixth fix, describing a further edge case in the same state machine:

> When the browser takes over a touch pan it emits `pointercancel`, which clears `isInteracting`; if the user then pauses their finger for more than 100ms without lifting, no scroll events refresh the native-scroll-end debounce, so it clears `isNativeScrolling` and the alignment effect programmatically snaps back to the nearest active slide while the gesture may still be live. Cancellation now starts the debounce (from an earlier fix in the PR), but the fixed 100ms timeout can still expire before the cancelled gesture actually ends.

## Ask

This is a real, if narrow, edge case: a touch pan that gets cancelled by the browser (native gesture takeover — e.g. a scroll-view lock, a system back-swipe, etc.) and where the user then pauses mid-gesture for >100ms without lifting their finger. The fixed-duration debounce in `scheduleNativeScrollEnd` can't distinguish "gesture ended" from "gesture paused," so the carousel may briefly reassert its own idea of the active slide while the browser still owns the track.

Options to evaluate, rather than another point patch:
- Track pointer-up/pointer-cancel-settlement more precisely instead of (or in addition to) a fixed timeout, e.g. don't run the realignment effect while `activePointerIds.size > 0` even after a `pointercancel`.
- Confirm real-browser behavior for `pointercancel` mid-pan (does another `pointerdown`/`pointermove` sequence normally follow, or is the gesture truly over) before deciding whether this is worth the added complexity.

## Why filed instead of fixed in #1010

The state machine (`isInteracting` / `isNativeScrolling` / `programmaticTarget` / `activePointerIds`) has now had six edge-case fixes in one review cycle, each surfaced by the previous fix. Per repo convention (`CLAUDE.md`: "map the edge-case matrix up front... a state machine designed once beats one patched across a dozen follow-up commits"), the remaining gap should be addressed as a deliberate redesign of the pointer/scroll ownership model, not as an eighth incremental patch riding an already-long review cycle.

Severity: P2/cosmetic (a momentary realignment snap), requiring a fairly specific real-world sequence (browser-cancelled touch pan + a paused, not-yet-lifted finger). Not a blocker for #1010.

stevekinney:
## Summary
- retain native-scroll ownership when the browser cancels a touch or pen pointer while the finger remains down
- settle the cancelled gesture only after the pointer is released
- add focused regression coverage and a patch changeset
- rebase exact head fe2c890b09a212e0ba8b5f46837bca3e81fe531e onto current main 389700023af97651b11ff4bf1d21962a935a76ba

## Verification
- `TZ=UTC LANG=en_US.UTF-8 bun test --conditions browser --conditions svelte --parallel=1 packages/components/src/components/carousel` — 47 pass
- `bun run --filter=@lostgradient/cinder components:check`
- `bun run --filter=@lostgradient/cinder lint:invariants`
- `bun run --filter=@lostgradient/cinder typecheck` — 0 errors
- `bunx prettier --check packages/components/src/components/carousel .changeset`
- pre-push fast sanity checks

Closes #1053

### DIFF
--- a/.changeset/carousel-pointer-cancel.md
+++ b/.changeset/carousel-pointer-cancel.md
@@ -1,0 +1,5 @@
+---
+'@lostgradient/cinder': patch
+---
+
+Keep Carousel native-scroll ownership until a cancelled touch or pen pointer is released.
--- a/.changeset/carousel-pointer-cancel.md
+++ b/.changeset/carousel-pointer-cancel.md
@@ -2,4 +2,4 @@
 '@lostgradient/cinder': patch
 ---
 
-Keep Carousel native-scroll ownership until a cancelled touch or pen pointer is released.
+Settle Carousel native-scroll ownership from the debounce after a touch or pen pointer is cancelled.
--- a/packages/components/src/components/carousel/carousel.svelte
+++ b/packages/components/src/components/carousel/carousel.svelte
@@ -60,7 +60,6 @@
   let viewportElement = $state<HTMLElement | null>(null);
   let programmaticTarget: number | null = null;
   const activePointerIds = new SvelteSet<number>();
-  const cancelledPointerIds = new SvelteSet<number>();
   let nativeScrollEndTimer: ReturnType<typeof setTimeout> | null = null;
   let scrollFrame: number | null = null;
   let cachedViewportInlineSize = 0;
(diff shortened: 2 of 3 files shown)


stevekinney:
@codex review

stevekinney:
@codex review exact head fe2c890b09a212e0ba8b5f46837bca3e81fe531e

PR #1067, opened 2026-07-31
fix(selection-popover): distinguish composer focus from toolbar focus

### LINKED ISSUE #1052: SelectionPopover: audit virtual-keyboard ownership latch for focus/ownership edge cases
## Context

PR #981 (closes #933) went through six rounds of automated review focused entirely on `SelectionPopover`'s virtual-keyboard ownership-latch machinery in `packages/components/src/components/selection-popover/selection-popover.svelte` (the `readVirtualKeyboardTransition` / `virtualKeyboardOwnedByComposer` / `keyboardOwnershipForgotten` logic, plus the `dismiss()` / `dismissVisualViewport()` guards that consult it).

Five of those rounds identified confirmed, fixed bugs:
1. `requestClose()`'s dedup latch only released on the next open transition, permanently swallowing dismissals when a controlled consumer declined a close.
2. `restoreFocus()` moving focus to a real pre-open owner fired a `focusin` that was mistaken for the user moving on, clearing keyboard ownership Cancel/Escape needed preserved.
3. The ownership latch was unconditionally overwritten by the *current* composer state on every event while the keyboard was still visible, instead of only ever latching to `true`.
4. Forgetting ownership on external focus only deleted the latch dict entry, so a stale `expanded`/draft state could immediately re-claim it on the next event.
5. The forgotten flag was never cleared when focus genuinely returned to the composer, leaving it stuck forgotten for the rest of the keyboard session.

The sixth (https://github.com/stevekinney/cinder/pull/981#discussion_r3671420493, unresolved-until-triaged) proposes requiring `ownedByComposer` even when `composerHasFocus` is true — i.e. removing `composerHasFocus` as an independent sufficient condition for suppressing a dismissal. That directly conflicts with existing, intentional test coverage (e.g. "a keyboard-driven visual viewport scroll preserves the focused draft") where `composerHasFocus` alone — with no prior keyboard-visible event to establish `ownedByComposer` — is relied on to protect an in-progress draft. Applying the suggestion as written would regress that guarantee.

## Ask

This mechanism has accumulated five successive patches to a single function's ownership bookkeeping across one review cycle. That's a sign the state machine (visible/hidden × owned/forgotten × composer-focused, across two independent event sources) deserves a from-scratch audit and possibly a redesign (e.g. an explicit state enum instead of three separate boolean/set variables threaded through closures), rather than another one-off patch layered on top blind.

Suggested next steps:
- Enumerate the full state space for `readVirtualKeyboardTransition` (visible/hidden × forgotten/not × composer-owns-now/not) and verify each transition against the existing test suite in `selection-popover.test.ts`.
- Specifically evaluate discussion #3671420493's scenario (external input owns an already-visible keyboard, then focus moves to the collapsed "Add comment" button via hardware-keyboard/switch navigation) and design a fix that doesn't regress the focused-draft-preservation tests.
- Ideally validate on a real device/OS keyboard (iOS Safari, Android Chrome) — this logic has never been exercised outside of `navigator.virtualKeyboard` / `visualViewport` mocks in `bun:test`.

## Non-goals

Not asking for a full rewrite as a blocker — `SelectionPopover` shipped with `#933`'s fix and five confirmed bugs already resolved. This is about scheduling deliberate follow-up work on the remaining edge cases rather than reacting to each new automated-review finding in isolation.

stevekinney:
Closes #1052

## Summary
- treat only the inline textarea as composer focus for virtual-keyboard movement guards
- dismiss external-keyboard movement after focus lands on the collapsed Add comment action
- add regression coverage and document the focus/keyboard accessibility contract

## Validation
- `git diff --check`
- focused Bun test blocked: assigned worktree dependency installation could not complete (`svelte/compiler` unavailable)

### DIFF
--- a/.changeset/selection-popover-keyboard-focus.md
+++ b/.changeset/selection-popover-keyboard-focus.md
@@ -1,0 +1,5 @@
+---
+'@lostgradient/cinder': patch
+---
+
+Dismiss SelectionPopover movement caused by an external keyboard after focus lands on the collapsed action.
--- a/packages/components/src/components/selection-popover/selection-popover.a11y.md
+++ b/packages/components/src/components/selection-popover/selection-popover.a11y.md
@@ -20,6 +20,13 @@ Purpose: Floating toolbar anchored to a text selection that exposes a comment-on
 
 The trigger should remain reachable by keyboard, and the open layer should provide a clear Escape, close, or outside-interaction path consistent with the component documentation.
 
+When the inline composer is expanded, virtual-keyboard movement is treated as an
+internal interaction only while the textarea itself owns focus. The collapsed
+“Add comment” control is not the composer: if focus arrives there while a
+keyboard belongs to another field, viewport movement may dismiss the popover.
+This keeps keyboard navigation from preserving a stale selection layer while
+still preserving drafts during genuine textarea-focused keyboard pan/resize.
+
 Keep focus indicators visible. If you wrap or restyle SelectionPopover, verify the focused element remains visually apparent in default and forced-colors modes.
 
 ## Names, roles, and state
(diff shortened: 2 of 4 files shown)


stevekinney:
@codex review

PR #1132, opened 2026-08-04
fix(text-direction): resolve native CSS nesting selectors

### LINKED ISSUE #1054: text-direction: resolve native CSS nesting (&) in direction-styling-hint selectors
## Context

Flagged by review on #993 (`packages/components/src/_internal/text-direction.ts`, `matchesDirectionStyleRuleList`).

When a direction override is authored with native CSS nesting, e.g.:

```css
.shell {
  & .cinder-mega-menu {
    direction: ltr;
  }
}
```

`matchesDirectionStyleRuleList()` passes the nested rule's `selectorText` (which
can retain the authored `&` token, depending on the engine's CSSOM
serialization) directly to `element.matches()`. `&` is not valid as a
standalone token for `Element.matches()`, so the match silently fails (a
`try`/`catch` swallows the `DOMException` and treats the rule as
non-matching).

## Impact

Under an RTL `LocaleProvider` with an ancestor whose *effective* computed
direction is LTR via a nested rule like the one above, `resolveTextDirection()`
fails to recognize the styling hint, keeps resolving RTL, and the affected
component (e.g. `MegaMenu`) runs its keyboard traversal and motion in the
opposite direction from the rendered (LTR) layout.

## Suggested fix

When encountering a nested `CSSStyleRule`, resolve `&` against the ancestor
rule chain (`rule.parentRule`) before calling `element.matches()` — walking
up `parentRule` to build the fully-qualified selector, recursively handling
multiple `&` nesting levels. This is a real (if narrow) parser addition, not a
one-line change, so it's tracked separately from #993 rather than bundled into
that PR's merge.

## Scope

- `packages/components/src/_internal/text-direction.ts`
- Needs unit coverage for: single-level nesting, multi-level nesting, and `&`
  appearing mid-selector (e.g. `.parent:has(&)`).

stevekinney:
Closes #1054

Resolve native CSS nesting `&` selectors against their ancestor CSSStyleRule chain before calling `Element.matches()`. Handles multiple nesting levels, selector lists, implicit nested selectors, and mid-selector references while preserving escaped syntax. Adds focused regression coverage.

### DIFF
--- a/.changeset/nested-direction-selectors.md
+++ b/.changeset/nested-direction-selectors.md
@@ -1,0 +1,5 @@
+---
+'@lostgradient/cinder': patch
+---
+
+Resolve direction overrides authored with native CSS nesting selectors.
--- a/packages/components/src/_internal/text-direction-css.ts
+++ b/packages/components/src/_internal/text-direction-css.ts
@@ -207,11 +207,20 @@ function splitSelectorList(selectorText: string): string[] {
   const selectors: string[] = [];
   let parenthesesDepth = 0;
   let bracketDepth = 0;
+  let quote: '"' | "'" | undefined;
   let start = 0;
   for (let index = 0; index < selectorText.length; index += 1) {
     const character = selectorText[index];
     if (character === '\\') {
       index += 1;
+      continue;
+    }
+    if (quote !== undefined) {
+      if (character === quote) quote = undefined;
+      continue;
+    }
+    if (character === '"' || character === "'") {
+      quote = character;
       continue;
     }
     if (character === '(') parenthesesDepth += 1;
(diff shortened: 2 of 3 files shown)


stevekinney:
@copilot please review the current exact PR head 2745eff00522308f8f5d1cd074d373619ef9a026 and leave inline comments for any actionable findings.

stevekinney:
@codex please review the current exact PR head 2745eff00522308f8f5d1cd074d373619ef9a026 and leave inline comments for any actionable findings.

stevekinney:
@copilot please re-review exact current head b21a5a660 after the quoted-selector parsing fix.

stevekinney:
@codex please re-review exact current head b21a5a660 after the quoted-selector parsing fix.

Issue #919, opened 2026-07-25
Epic: components hand-roll primitives instead of composing them — 10+ confirmed, and the copies drift


stevekinney:
## Problem

A design system's core promise is one implementation per primitive. Cinder currently has many. Verified duplications, each with file:line:

**Form controls that do NOT import `Input`:**
- `autocomplete/autocomplete.svelte` renders a raw `<input class="cinder-autocomplete__input">` at `:447-452`, styled at `autocomplete.css:17-26`, drifted from `input.css:41-58` on background, padding, min-height, font-size, line-height, transition, and `::placeholder`.
- `number-input/number-input.svelte`
- `phone-input/phone-input.svelte`
- `search-field/search-field.svelte:121`
- `time-field/time-field.svelte`
- `event-stream-viewer/event-stream-viewer.svelte:323-330` — raw filter `<input>`; the component only imports `CopyButton`, `JsonViewer`, `StatusDot`.

**MultiSelect reimplements Checkbox and Combobox:**
- `multi-select/multi-select.svelte:511-513` renders a `<span>` with a Unicode `✓` and `aria-hidden`, instead of `Checkbox` (`checkbox.svelte:132` uses a real `<input type="checkbox">`).
- Its dropdown (`multi-select.svelte:449-533`) reimplements open/close, outside-click and focus-out dismissal (`:356-387`), roving `activeIndex`, and scroll-into-view (`:389-393`) — while `combobox.svelte:426,437` independently hand-rolls the same listbox pattern.

**Every labelled control renders its own label/description/error wrapper instead of composing `FormField`**, whose stated purpose is exactly that:
- `Input` renders `.cinder-input-field` (`input.svelte:157`).
- `Toggle` renders `.cinder-toggle-field` (`toggle.svelte:55`).
- **This is the root cause of the `fieldClass` wart** (`checkbox.types.ts:35`) — Checkbox is the only one that exposed its wrapper, so it invented a one-off prop name.

**Grid duplications:**
- `bento-grid` — near-identical prop surface and CSS to `grid`, differing only in default columns, `minmax(0,1fr)` sizing, and a `collapse` prop, while LACKING `minItemWidth`.
- `grid-list/grid-list.css:17-21` hard-codes `repeat(auto-fill, minmax(min(var(--cinder-grid-list-min-width),100%),1fr))` which is exactly `Grid`'s `minItemWidth` (`grid.types.ts:24`).
- `logo-cloud`
- `choice-grid`

[...]

PR #1144, opened 2026-08-05
refactor(components): compose FormField across remaining field wrappers

stevekinney:
## Summary

Finishes the field-wrapper composition sweep (epic #919). Every remaining entry in `allowedFieldWrapperCounts` is now migrated to compose `FormFieldFrame` for label/description/error, following #1120's pattern (Input/Checkbox/Toggle). No entries needed a written structural-exception justification — all eight migrated cleanly.

## Per-component outcomes

- **`_radio/radio.svelte`** — migrated. The row is now `FormFieldFrame`'s own root (merged with `.cinder-form-field`). Retired `data-checked`/`data-disabled`/`data-invalid`/`data-has-description` attributes became BEM modifier classes (`radio-group.css` selectors, `radio-group.a11y.md`, `radio-group.test.ts` updated to match). `order` CSS restores the original control-first visual layout since FormFieldFrame's fixed render order is label→control.
- **`combobox/combobox.svelte`** — migrated. Listbox/roving-index state untouched, per the caution in scope — only the field wrapper (label/description/error) moved to `FormFieldFrame`, using the new `errorAlwaysMounted` + explicit `errorId` override to preserve the pre-mounted live-region behavior.
- **`date-picker/date-picker.svelte`** (2 tracked wrappers) — migrated. Root label/description/error moved to `FormFieldFrame`; the embedded "Time" sub-label (for the granularity time input) stays native markup — it has no description/error pairing of its own, so it isn't a field-wrapper pattern. `date-picker.css`'s root selector (`display: grid`) genuinely conflicted with `.cinder-form-field`'s base `display: flex` — date-picker.css imports *before* form-field.css in the aggregator, so this was a real, proven cascade bug, not speculative; fixed with a compound `.cinder-date-picker.cinder-form-field` selector so correctness doesn't depend on import order.
- **`json-editor/json-editor.svelte`** — migrated. `feedbackText` (JSON parse status, either an error or a "Valid JSON." success message — role="alert" or role="status") is richer than FormFieldFrame's plain `error` slot, so it's routed through the new `message` snippet instead. `json-editor.css`'s root `gap` also conflicted with form-field.css's base gap value; fixed with the same compound-selector pattern (json-editor.css already imports after form-field.css, so this was latent, not live, but the value conflict is real). `order` CSS restores the original label→description→input→feedback visual order.
- **`multi-select/multi-select.svelte`** — migrated (field wrapper only; the tracked raw-control exception for its filter input and native validation proxy is untouched, per its own allow-list entry). `warning` merged into FormFieldFrame's `message` slot; `error`/`nativeError` merged into a single `error` prop.
- **`select/select.svelte`** — migrated. Same `errorAlwaysMounted` pattern as Combobox.
- **`textarea/textarea.svelte`** — migrated, with `{#if context}` branching (matching Checkbox's established #1120 pattern) so nesting Textarea inside an external `<FormField>` with no local label/description/error doesn't render a redundant empty wrapper div, and so a local label isn't rendered a second time when the outer FormField already supplies one.
- **`time-field/time-field.svelte`** — migrated. Found and fixed a real regression in the inherited WIP: `{...rest}` (typed `HTMLAttributes<HTMLDivElement>`, meant for the field root) had been relocated onto the inner `.cinder-time-field__controls` row instead of following the migration to the root. Fixed by giving `FormFieldFrame` itself a generic root-attribute passthrough (additive, empty for every other caller) and restoring `{...rest}` there. Added a regression test asserting forwarded attributes land on `.cinder-time-field`, not the inner row (and the same assertion for DatePicker, which composes correctly today but had no test proving it).

## FormFieldFrame changes

Added `labelClass`, `errorAlwaysMounted` (pre-mounts the error live region so screen readers pick it up before text is injected — needed by Select/Combobox/MultiSelect), a `message` snippet slot (extra content between description and error — JsonEditor's status text, MultiSelect's warning, Textarea's character counter), and generic HTML attribute passthrough on the root `<div>`.

## Cascade-safety verification

Checked every migrated component's CSS root/description/error rules against `form-field.css`'s base rules for genuine value conflicts (not just import order) across the aggregator (`src/styles/components.css`). Only `date-picker.css` and `json-editor.css` had real conflicts (documented above); both fixed with `#1120`'s established compound-specificity pattern (already used by `checkbox.css`), which makes correctness independent of aggregator import order rather than adding an order-pinning regression test — there's nothing left that's load-bearing on order to pin.

## Allow-list

`allowedFieldWrapperCounts` is now an empty `Map` (kept, not deleted, so a future hand-rolled field wrapper still trips the guard). `check-primitive-composition.test.ts` updated: the "drops completed migrations" test now covers all eight paths, and the generic "tracks occurrences" test was rewritten against a synthetic path (via `try`/`finally` `set`/`delete`, matching the existing `allowedRawControlSignatures` precedent) since it no longer had a real tracked entry to exercise the mechanism against.

## Validation

- `bun run check:primitive-composition` — 0 stale entries, 0 violations
- `bun run typecheck` (tsc + svelte-check) — 0 errors, clean across 3 consecutive runs
- `bun run lint:invariants` — all 13 checks pass
- `bun run --filter=@lostgradient/cinder components:check` — OK
- `bun run lint` (oxlint, full package, 1500 files, 107 rules) — 0 warnings/errors
- `prettier --check` on all touched files — clean
- Per-component `bun run test <name>` for every touched/dependent component (radio-group, combobox, date-picker, json-editor, multi-select, select, textarea, time-field, form-field, input, checkbox, toggle) — all green
- `bun run test scripts` — 2692/2692 pass
- `changeset status` — `@lostgradient/cinder` recognized correctly (aggregated into the existing pending minor bump from other unreleased changesets)
- Playwright, run locally against a clean 5555/5556 window: date-picker surface-ownership spec (1/1), Combobox visual snapshots plus the floating-surface-containment spec that clips `.cinder-combobox` (7/7), and a sweep across Select, MultiSelect, Textarea, TimeField, JsonEditor, RadioGroup, and everything else the grep matched (82/82, 0 axe violations)

## Known non-blockers (not from this change, reported not chased)

- A whole-package `bun run test` run intermittently fails on `approval-card.test.ts` timing out and unrelated `share-card`/OOM signal-kill under heavy concurrent-worktree load on this machine — reproduced with my changes fully stashed too, so it's environmental, not this PR. Every file this PR touches has clean, repeated, deterministic per-component and per-script test runs.
- `svelte-check` intermittently (once, not reproducible) flagged `bento-grid/bento-grid.svelte` and `speed-dial-action/speed-dial-action.svelte` — both untouched by this PR — with unrelated errors; three consecutive clean re-runs on the same final tree showed 0 errors. Flagging in case it recurs elsewhere; not fixed here since these files are out of scope.

Epic: #919


stevekinney:
Follow-up push fixed two issues:

1. **Copilot review** — `data-cinder-error` now reflects `error` presence whenever the error node is rendered, not just when `errorAlwaysMounted` is set.
2. **A real bug caught by CI's `unit-tests` job**, not this session's local runs: `message={condition ? snippet : undefined}` does not reliably re-toggle `FormFieldFrame`'s `{#if message}` when `condition` changes reactively — the stale snippet content stays mounted. Reproduced directly against JsonEditor (confirmed via `ApprovalCard`'s "edits arguments as JSON" test), then found the identical latent pattern in MultiSelect's `warning` and Textarea's `count`. Fixed all three by passing the snippet reference unconditionally and guarding the content with an internal `{#if}` inside the snippet body instead. Added a regression test per component (JsonEditor via a real `fireEvent.input` edit, MultiSelect and Textarea via `rerender`) — each fails against the old ternary pattern and passes against the fix.

Worth flagging for future FormField-composition work: **don't conditionally pass a snippet prop as `condition ? snippet : undefined`** — pass the snippet unconditionally and gate its content internally instead.

Also merged `main` twice to pick up #1140 (docs) and #1141 (fix(publish): ship `src/_internal` Svelte components in the tarball — directly relevant since `form-field-frame.svelte` lives there). No conflicts either time.

PR #1247, opened 2026-08-10
fix(chat): settle virtualized bottom pin after measurement

### LINKED ISSUE #1243: Chat: virtualized bottom pin oscillates at rest — scrollHeight flips continuously and the auto-stick effect re-fires in a loop
## Observed

Found while diagnosing #1236 (instrumented real-browser trace against the playground harness's 300-message varied-height virtualized thread, `@lostgradient/chat` main as of PR #1242): with the transcript simply resting pinned at the bottom — no user input, no streaming — the virtualizer's `scrollHeight` flips back and forth between two values (23826 ↔ 23762 in the traced run) and the auto-stick-to-bottom `$effect.pre` re-fires continuously, issuing a fresh instant `scrollToOffset(scrollSize)` correction on every flip.

## Why it matters

- It's a standing busy-loop of layout + scroll work on an idle screen (measure → height flip → auto-stick correction → measure …).
- The steady stream of in-flight corrections is exactly the stale scroll/scrollend generator that broke `scrollToTop()` in #1236 — PR #1242 hardens the guard against those stale events, but the generator itself remains. Any future feature that trusts "the viewport is quiet at rest" will trip over it.

## Suspicion

Two sources of row measurement disagree (~64px, roughly one row): the estimated size for a not-yet-measured row vs its measured height, or a measurement taken mid-correction — so each correction invalidates the measurement that justified the previous one. Untraced beyond the flip values; the oscillation itself reproduces on the playground harness's `seed-long` control added in PR #1242.

## Requested

Make the resting state actually rest: converge `scrollSize` after measurement (no flip-flop between estimated and measured heights), and/or make the auto-stick effect no-op when the delta is below a row-measurement epsilon instead of issuing a correction per remeasurement.

stevekinney:
Fixes #1243

The auto-stick effect now checks the live viewport distance from the maximum offset before issuing an instant correction. This prevents repeated corrections when virtual row measurement re-runs after the viewport has already settled at the bottom.

Tests: focused Chat virtualization suite (20 pass).

@copilot please review
@codex please review

### DIFF
--- a/.changeset/quiet-bottom-pin.md
+++ b/.changeset/quiet-bottom-pin.md
@@ -1,0 +1,5 @@
+---
+'@lostgradient/chat': patch
+---
+
+Avoid repeated virtualized bottom corrections when the transcript is already settled at the bottom after row measurement.
--- a/packages/chat/src/lib/components/chat/chat-virtualization.test.ts
+++ b/packages/chat/src/lib/components/chat/chat-virtualization.test.ts
@@ -295,6 +295,36 @@ describe('Chat virtualization', () => {
     }
   });
 
+  test('settled virtualized bottom does not issue repeated corrections after measurement', async () => {
+    let conversation = longConversation(20);
+    const { container, rerender } = render(Chat, {
+      props: virtualizedProps(conversation),
+    });
+    const timeline = await waitForVirtualizedTimeline(container);
+    Object.defineProperty(timeline, 'clientHeight', { configurable: true, value: 100 });
+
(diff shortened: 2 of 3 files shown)


stevekinney:
@copilot please review this pull request. @codex please review this pull request.

stevekinney:
@copilot please re-review the latest head 15d2e211. @codex please re-review the latest head 15d2e211.

stevekinney:
@copilot review

@codex review

stevekinney:
@copilot review

@codex review

Issue #1284, opened 2026-08-12
editor: cut-and-paste of commented text deletes the thread permanently once the re-anchor debounce elapses


stevekinney:
Data loss in normal editing. Reported against `@lostgradient/editor@0.8.1`, and I want to be upfront that **this became reachable because of a fix I made** in #1266.

## What happens

Cut a paragraph containing an anchored comment, pause for more than ~300ms, paste it back. The text returns; the thread does not. There is no undo path — it is gone from the bindable `threads` array and `onthreaddelete` has already fired.

The consumer's suite covers it today:

```ts
// `drift-move-slow` reinserts the deleted word 450ms after deleting it.
const LATE_PASTE_SETTLE_MS = 1400;
…
await page.getByTestId('drift-move-slow').click();
await expect(page.getByTestId('drift-thread-count')).toHaveText('threads: 0');
await page.waitForTimeout(LATE_PASTE_SETTLE_MS);
// The text is back, exactly where the successful burst put it.
await expect(page.locator('#anchor-drift p')).toHaveText(/^dashboard The first release/);
// The thread is not.
```

The same move performed inside one synchronous burst re-anchors correctly and writes back to `threads`. The only variable is human-speed latency.

## Mechanism

`anchor-decorations.ts` debounces re-anchoring 300ms after the last document change. On a cut, the anchor's range collapses, `reanchorQuote` searches a document that no longer contains the quote, returns `found: false`, and the thread is queued for deletion. The module comment says collapsing is deliberate so that "cut/paste and move operations where the text may reappear elsewhere" survive — but the window is 300ms, and a person cutting and pasting is routinely slower than that.

Before #1266 the consequence was invisible: `ReviewEditor` constructed the plugin without an `onAnchorDeleted` handler, so detection fired into nothing. The thread stayed in `threads` with a stale anchor. That was its own bug — the documented contract in `comments/types.ts` says "When anchor text is deleted, threads are automatically removed", and it did not happen — so #1266 wired the handler.

Wiring it was right for a genuine deletion and wrong for a cut. The two are indistinguishable at 300ms, and the fix converted a silent inconsistency into silent data loss. That is a worse failure mode, and it is on me.

## Why a longer debounce is not the fix

Any fixed window still splits real deletions from slow cut/paste arbitrarily, and it delays the legitimate case by however long it is set to. This is not a timing problem to tune.

## Suggested direction

[...]

PR #1285, opened 2026-08-12
fix(editor,chat): orphan vanished anchors, free the Tab trap, and five more

stevekinney:
Six defects found by exercising both components as a consumer. Every fix carries a regression test **verified to fail without it** — reverted locally and observed red, not written to match new behavior.

## editor — a cut-and-paste destroyed the comment (#1284)

The one I would not ship without. Re-anchoring is debounced 300ms, and deletion is indistinguishable from a cut at the moment the text disappears. Cut a commented paragraph, pause, paste it back: the comment was gone, no undo.

**This became live because of my own fix in #1266.** `comments/types.ts` documented that threads are deleted when their anchor text goes, and the component never wired the handler, so it silently did not happen. Wiring it honoured the contract — correct for a real deletion, wrong for a cut — and turned a silent inconsistency into silent data loss.

So the contract changed instead. `AnchorStatus` gains `'orphaned'`: a vanished quote marks the anchor orphaned and **keeps** the thread, which renders no decoration and is retried on every later pass, so restoring the text restores the anchor. The sidebar marks it `[data-orphaned]` with "Quoted text is not in the document". Removing a thread is now the consumer's decision; `onAnchorDeleted` is deprecated and never called.

The consumer's `drift-move-slow` fixture — delete, wait 450ms, re-insert — now asserts **recovery**, byte-identical to the same move performed in one burst.

## editor — Tab was a keyboard trap in list items (WCAG 2.1.2)

Tabbing in puts the caret at the end of the document. If that block was a list item, the sink/lift keymap consumed Tab and Shift+Tab, so the only way out was to keep re-indenting the bullet — and whether you were trapped depended on a block boundary you cannot see. Tab still indents inside a list; there is now an escape.

## editor — `generateUnifiedDiff` corrupted YAML front matter

`normalize()` re-read the `---` fences as a thematic break plus setext heading, injecting 8-dash lines and wrong hunk headers. Verified by `git apply --check` rejecting the output, while the docs promise appliable patches. It shipped through `exportUnifiedDiff()`, the `<name>-diff` hidden input, and Copy Diff. Front matter is now parsed off, only the body normalized, and the front matter re-attached verbatim — matching what DiffViewer already did.

## editor — `deleteComment` silently no-opped on the event the component emits

It required `options.deletedAt`, but `CommentDeleteEvent` carries only `{ threadId, commentId, soft }`, and the option is typed optional — so the obvious wiring typechecked and did nothing, after "Comment deleted" had already been announced to screen readers. The reducer now stamps the timestamp itself; an explicit one still wins.

## editor — the loading placeholder was visible text

`EditorSkeleton` hid "Loading editor..." with a bare `sr-only` class Cinder does not ship, so it rendered as body copy during load and **permanently** for a no-JS reader. Same defect already fixed in `review-editor/live-region.svelte`.

## chat — jump-to-latest dropped focus onto `<body>`

In a virtualized transcript, focusing a message row one tick after scrolling focuses a recycled window slot; the virtualizer's next pass unmounts it and focus falls to `<body>`. Since the keydown handler is bound on the container, that killed **every** shortcut — End, Home, PageUp/PageDown, arrows, Ctrl+F — on the configuration recommended for long histories, through three releases, because no test pressed End.

The selector was independently wrong: `.chat-message-wrapper:last-of-type` matched every wrapper (each is the only child of its virtual row), so it focused the row at the **top** of the window. Home had the identical defect. Both now focus the stable `.chat-timeline`, with a `focusout` backstop for a row unmounted from under the user.

## Verification

- editor **626 pass**, chat **774 pass**, 0 failures; typecheck and lint clean on both
- consumer suite **296 passing** against a local build of both packages
- six consumer pins that asserted the old behavior were rewritten to assert the new contract, keeping the history in their comments
- one genuinely stale pin retired: `review-ssr-and-a11y` still claimed the control bar was `role="toolbar"` (it has been `role="group"` since 0.8.1, which the same file already asserted 200 lines later) and passed only because it counted tab stops

## Not included

The audit that produced this list also flagged: per-keystroke diff cost under `name`, a dual normalizer between toolbar and diff panel, `reset`/`setMarkdown`/`setState` skipping the readonly guard, DiffViewer's window-level key bindings, three quiet-failure paths in the sidebar, and a slot-semantics divergence — Chat's overrides pass `renderDefault` while DiffViewer's `toolbar` is total replacement, which is why ReviewEditor's empty snippet deletes the Compute Diff button. Each is real; none blocks a consumer who is not in that configuration. Happy to take them next.

https://claude.ai/code/session_01HhsXZt1sQAX5bNu3Pv4DuN


stevekinney:
## Review round 2 — all five findings addressed

### P1: the `focusout` backstop could not fire for the case it existed for — **correct, fixed**

This was the real one, and the criticism of the test was exactly right: dispatching a synthetic `FocusEvent` before `remove()` made the test prove the handler works when called, not that anything would ever call it.

The backstop is now scroll-driven. `focusin`/`focusout` still track *which* row holds focus, but they no longer decide that a reclaim is due — `reclaimFocusIfRowDetached()` re-checks that row's `isConnected` on each scroll-state recompute, since scroll is the only thing that recycles rows. (A `MutationObserver` also catches it and is what this originally did, but `subtree: true` on a list that mutates every scroll frame is not worth it for a check this cheap — and in this harness it drove Bun's RSS to 8.8GB before segfaulting.)

**I removed the isolated recycling test rather than make it green.** Driving the honest version — detach a row, let a real scroll recompute notice — segfaults the happy-dom harness in three different formulations: hand-removing a node the virtualizer believes it owns and then running its measurement pass over the result is not a state it survives, and the `waitFor` polling the rAF-deferred recompute needs makes it worse. The file now says so explicitly, in place of the test.

What still has coverage: the user-facing consequence (after a real `End`-driven window pass recycles the rows, keyboard shortcuts still work) and the negative case (clicking away onto inert page chrome does *not* trigger a reclaim). The gap is the isolated unit, and it wants a real browser — chatroom's Playwright suite.

### P2: document-level anchors orphaned on every edit — **not reproducible; I believe this one is incorrect**

Document-level anchors (`type: 'document'`) legitimately carry `from: 0, to: 0` and an empty quote, so the concern is reasonable on its face. But they never reach the quote-search path: `computeDecorations` and the re-anchoring pass both skip non-`text` anchors before any lookup happens, so there is no `!result.found` to turn into an orphan.

I did not want to take that from reading alone, so I added a permanent assertion covering a document-level thread across edits before, inside, and after its nominal range. 19/19 pass, including the new one, and it fails if the type guard is removed. Happy to look again if there's a specific sequence I missed.

### The other three — fixed as described

- **`setState` dropped orphans**: it rebuilt anchors without preserving `status`, silently resurrecting orphaned threads as anchored at `0,0`. Now preserved.
- **Sync fingerprint omitted `status`**: a thread that flipped anchored→orphaned with unchanged positions did not re-render the sidebar. `status` is in the fingerprint.
- **`coordsAtPos` on an orphaned anchor**: `0,0` is a valid position, so this threw no error — it silently placed the popover at the top of the document. New `anchorCoords()` guard returns null for orphans, and the sidebar row falls back to its own rect.

### Verification

`editor` 626 pass / 1 skip / 0 fail · `chat` 774 pass / 0 fail · both typecheck clean · repo `lint` clean (3 pre-existing warnings, 0 errors).

Two doc comments and the `chat-jump-focus` changeset described the reverted `focusout`/microtask design; both are rewritten to match what actually ships.

stevekinney:
## Review round 3

Three of the new findings were verified independently, each by writing a test and then checking that test is load-bearing. Two do not reproduce; one was real and is fixed.

### Not reproducible: "Clear orphaned status when the quote already matches" (P1)

Does not reproduce, for re-insertion at the same position **or** a different one — the status is cleared back to `anchored` in both cases.

The claim describes the code at `992ee8204`. It was fixed one commit later in `5a2409dc0`, whose message describes this exact failure in nearly the same words. The clearing happens inside the very `anchorMatchesDocument` fast path the comment names (`anchor-decorations.ts:522-538`): when the mapped anchor's quote matches again and the status is `orphaned`, it emits an update with `status: 'anchored'`. `mapAnchorsThroughTransaction:377` is what schedules that pass on any doc change.

New permanent coverage in `anchor-decorations-orphan-recovery.test.ts`. It is not vacuous: run against the pre-fix module, the same-position test fails with `Expected: "anchored"  Received: "orphaned"`. Notably the different-position test passes even pre-fix, which is precisely why the original suite missed this.

### Not reproducible: "Stop continuously rescheduling unchanged orphaned anchors" (P1)

Also fixed in `5a2409dc0`. The guard is at `anchor-decorations.ts:266` — `if (anchorState.status !== 'orphaned' && !anchorMatchesDocument(doc, anchorState))` — so a preserved orphan no longer re-arms `needsReanchor`, and the plugin view's early return means no new timer.

Measured directly, idle document, one absent quote, ticking 300ms:

```
CURRENT CODE (guard present)          GUARD REMOVED
  t=  300ms: transactions=2             t=  300ms: transactions=2
  t= 3000ms: transactions=2             t= 3000ms: transactions=11
  t=60000ms: transactions=2             t=60000ms: transactions=201
```

Exactly the unbounded stream described — without the guard. Pinned in `anchor-decorations-orphan-loop.test.ts`, which counts armed timers against a controllable clock. The one legitimate retry (`:377`) fires once per document change, which is what recovery needs.

### Real and fixed: "Align the exported anchor manager with orphan preservation" (P2)

**This one was right, and it mattered more than P2.** `createAnchorManager` is a shipped, typed, importable API (`@lostgradient/editor/review-editor`, re-exported at `components/review-editor/index.ts:33`, present in `dist`), and it was the one remaining path that still deleted a thread whose quote had gone missing — the exact data loss #1284 is about.

Verification followed the same discipline: the tests were written first as characterization tests asserting the *old* behavior. After the fix, three flipped to failing and the found-quote control kept passing, which is what makes them meaningful now that they are rewritten to the contract.

Two adjacent gaps came out of reading it, both fixed here:

- **No `isDocumentAnchor` guard.** An empty quote can never be "found", so document-level threads were deleted despite not being lost. (This is the same concern as the earlier document-level P2 I pushed back on — I was right that the *plugin* path guards it, and wrong to treat the concern as closed, because this path did not.)
- **A silent drop.** When a quote resolved in the text but its offsets did not map to positions, the thread fell out of the loop unpushed — no orphan, no event, just gone.

`onthreaddelete` is removed from `AnchorManagerOptions` rather than left in place: it would never fire again, and a consumer wiring cleanup to an event that never arrives has no way to notice. The manager's sync fingerprint also gained `status`, matching the inline fix.

One pre-existing bug is documented, not fixed: this manager never parses front matter, so restored anchors are off by the front matter's length on any document that has some. It is now in the module docblock as the reason to finish or delete this factory rather than adopt it.

### Also in this round

- **Arrow navigation after virtualized Home** (P2) — real. Home focuses the viewport there, but the arrow handlers only fired on a focused `.chat-message`, so the fix for one keyboard dead end created another. Arrows now also enter from the viewport: ArrowDown at the first rendered message, ArrowUp at the last. Both new tests fail without the change.
- **Temp-directory leak in `unified-diff.test.ts`** — not as described; the `finally` with `rmSync(..., { recursive: true, force: true })` was already there, with a comment. There was a real narrower gap: `git init` and the writes sat outside the `try`, so a failure in setup leaked the directory. Moved inside.
- **Orphan announcement grammar** — the current strings contain no "They" and no possessive; that comment describes an earlier revision of the text. Left as-is.

### Verification

`editor` 636 pass / 1 skip / 0 fail (up 10) · `chat` 777 pass / 0 fail (up 3) · both typecheck clean · repo `lint` 10/10 tasks, 0 errors.

stevekinney:
## Round 4 — "Propagate status in the exported anchor manager" (P2): correct, fixed

Right, and it caught that my previous commit was half a fix. `handleAnchorsUpdate` copied every updated anchor field except `status`, so an anchor the plugin orphaned during live editing kept reaching consumers as `anchored` — unable to surface or persist the orphan while the plugin had already stopped decorating it.

Worth naming the sharper point in the comment: this made the fingerprint change in the same commit **inert**. Adding `status` to the sync fingerprint compares a value the manager never applied, so it could never differ. Two changes that each looked right in isolation and did nothing together.

Fixed by copying `update.status` onto the thread anchor, matching `review-editor-impl.svelte:490`. New test drives the live-editing route specifically (`handleAnchorsUpdate`, not the `setState` pass the other four cover); removing the one line makes it fail.

`editor` 638 pass / 1 skip / 0 fail · typecheck clean · repo lint 10/10 tasks, 0 errors.

stevekinney:
## Round 5 — three of four real, one not reachable

### Represent orphaned anchors in exported comments (P2): correct, fixed

Reproduced exactly. With one `orphaned` and one `anchored` thread carrying identical anchor data, **all three exports were byte-identical between them** — nothing distinguished the thread whose text is gone:

| Format | Orphaned thread, before |
|---|---|
| `generateCommentsJSON` | `"selection": { "text": "vanished text", "from": 50, "to": 63, "line": 3, "column": 11 }` |
| `generateCommentsExport` | `### Comment at Line 3:11` / `*Position: Line 3, Column 11 (offset 50)*` |
| `generateMarkdownSummary` | `### On "vanished text"` |

One correction: the summary never emitted a line number, so there the only misleading signal is the bare quote implying the text is present. The JSON and Markdown halves are exactly as described. Two things not mentioned that I checked: `generateUnifiedDiff` never touches threads, and `getFormData().comments` already carried `status` through `toPersistedThreads` — its `summary` did not. `generateCommentsJSON` had **zero test coverage** before this.

JSON now emits `status` and moves the stale offsets to `lastKnownSelection`; markdown and summary say the text is no longer in the document. `status` is emitted only when it is not `anchored`, so anchored output stays byte-for-byte identical — pinned by tests, including an exact `Object.keys` assertion. Revert proof: 44 pass / 8 fail with the sources stashed, 52 / 0 restored.

### Preserve orphan disambiguation offsets across retries (P2): correct, fixed

Reproduced, against live code at `6a0b110ab`. `anchor-decorations.ts:390` recomputed `lastKnownOffset` from the `0/0` sentinel, yielding `0` (`Expected: 448  Received: 0`).

The consequence is the one named. In a four-row document with byte-identical rows — so all four occurrences score 1.0 on context and only proximity separates them — restoring the fourth row's orphan and typing the text back gives `Expected: 452  Received: 59`: the comment reattaches to row **one**.

Probing adjacent paths found a second hole in the *drifted* branch: inserting a block at position 0 expands `0/0` (since `map(0,-1)=0` while `map(0,1)` moves past the new node), so the anchor never reaches the collapsed branch at all. Adding a paragraph at the top of a document is an ordinary edit. Both sites now guard on one shared unplaced-orphan check. Full revert: 3 fail / 1 pass; reverting only the drift hunk: 1 fail / 3 pass, isolating it as independently load-bearing. The test that passes in every configuration is the over-fix guard — an anchor collapsed at a *real* position still tracks the document.

Deliberately not done: mapping the preserved offset through every edit while unplaced. That is two full position-map builds per edit per orphan, and a stale-but-approximately-right hint beats that cost. Resetting to 0 was the defect.

### Clear tracked rows when focus deliberately leaves (P2): correct, fixed, untested

Real gap, and fixed — a pointer landing outside the container now clears the tracking, so a later recycle cannot haul focus back into a chat the user had left.

No test ships. The one I wrote passes identically with and without the fix, because happy-dom does not blur a focused element when a pointerdown lands on inert chrome. A test that cannot fail is worse than no test, so it is not in the diff; the reasoning is in the code instead.

### Reclaim focus after non-scroll row removals (P2): not reachable, and the reason is a separate bug

The mechanism is right, but the scenario cannot occur: **Chat does not remove message rows when the conversation shrinks at all.**

A Chat rendered with 6 messages and re-rendered with 3 still renders 6, container-scoped, with one instance mounted. Same for dropping a single id, and same for `hidden: true` — even though `getMessages` filters on `!message.hidden`. Meanwhile message *content* updates flow through normally on the identical path, so this is not a dead prop.

Filed as #1286. I have added the rendered-set effect anyway and labelled it honestly in the code: it cannot fire today, it costs one early-returning call per rendered-set change, and it is the difference between the backstop working and not the moment #1286 is fixed.

### Also

Two doc claims in `comments/types.ts` that this branch made false are corrected — the module header still said "When anchor text is deleted, threads are automatically removed", and `toRuntimeThreads` still promised `onthreaddelete` on a missing quote.

### Verification

`editor` 658 pass / 1 skip / 0 fail (up 20) · `chat` 777 pass / 0 fail · both typecheck clean · repo lint 10/10, 0 errors.

stevekinney:
## Round 6 — all three real. The P1 was a process failure on my side.

### Replace the added 450ms wait with deterministic timing (P1): correct, and it should not have shipped

This one is on me. The recovery tests slept a real 450ms per `settle()`, commented "plus slack for a loaded CI machine" — precisely what [AGENTS.md](https://github.com/stevekinney/cinder/blob/main/AGENTS.md#pull-request-review-test-timeouts-are-not-a-fix) forbids, in a repo whose own rule says "There is no exception." Several tests call `settle()` more than once, so the file was contributing seconds of nondeterministic wall clock by itself.

The right answer was already sitting in the adjacent files: `anchor-decorations-orphan-loop.test.ts` and `-offset.test.ts` each had a local `FakeClock`. Extracted it to `packages/editor/src/lib/test/fake-clock.ts`, pointed all three files at it, and deleted the two duplicates. It does not ship — `package.json` excludes `dist/**/test/**` and nothing outside tests imports it, same as the existing `test/happy-dom.ts`.

Still load-bearing after the conversion: breaking the recovery branch fails the test in **1.04ms** instead of after a sleep. The editor suite went from **4.12s to 1.83s**.

I also re-scanned the entire branch diff for any other added `timeout`, `retries`, `waitFor` deadline, or `slow()` multiplier. There are none. The only remaining `setTimeout(resolve, 0)` calls in the package are macrotask yields that predate this branch.

### Avoid inventing a last-known offset for legacy orphans (P2): correct, and understated

Reproduced, and there were **two** fabrications rather than one:

- **As reported**: no `lastKnownOffset` gave `lastKnownSelection: { from: 0, to: 13 }` via `?? 0`. Since the quote is by definition absent, a consumer has nothing to check it against — it just points at the top of the document.
- **Not reported, and worse**: an anchor with `originalPosition: {offset: 50, line: 3, column: 11}` but no `lastKnownOffset` exported `from: 0` *alongside* `line: 3, column: 11`. That describes no document that ever existed, and it discards a real historical offset — `comments/reanchor.ts:245` already uses `lastKnownOffset ?? originalPosition?.offset` for exactly this purpose.

Now falls back to `originalPosition.offset`, and omits `lastKnownSelection` entirely when neither exists. `status` is always emitted, so an orphan stays identifiable either way. Revert proof: 54 pass / 2 fail, restored 56 / 0.

Two claims in the finding I checked and did **not** act on, because they are not true of the current code: the markdown export appends `(offset N)` only when `lastKnownOffset !== undefined` and otherwise says only "This text was not found in the current document"; the summary prints no coordinates at all. Guard tests added so both stay that way.

### Clear the Tab escape latch when focus leaves (P2): correct, verified against a real EditorView

Reproduced. `release()` validated the latch with `armed.doc === state.doc && armed.selection.eq(state.selection)` — a proxy for "no transaction since Escape". A focus round trip applies no transaction: ProseMirror's blur/focus handlers only toggle `view.focused` and a class, and the DOM observer dispatches a selection transaction only when the read selection differs. Returning to the same caret leaves `view.state` identical **by reference**, so the latch stayed armed with no expiry.

Probe against a real view, pre-fix, focus moved to an `<input>` and back:

```
[real] blur listener fired: 1
[real] transactions: 0
[real] state identity: true
[real] Tab handled (true = indent): false   calls: []
```

Fixed with an arm-scoped `blur` listener on `view.dom`, disarmed on blur, on release, and on re-arm. Deliberately **not** `props.handleDOMEvents.blur`, though that does fire: registering a plugin from this module pushes into `prosePluginsCtx`, and `@milkdown/core`'s `editorState` snapshots that slice after `waitTimers(editorStateTimerCtx)` without waiting on user plugin runners — while this runner first awaits four dynamic imports. That is a real window where the registration lands too late and the fix silently does nothing.

Adjacent invalidations, checked: a typed character clears it, a caret move clears it, and a click landing **exactly** on the existing caret does not, because ProseMirror dispatches nothing when the read selection equals the current one. That residual survives, and is left alone on purpose: it needs a click hitting the caret to the character with the editor already focused, and covering it means disarming on `mousedown`, which is wider than this finding.

Revert proof: 12 pass / 4 fail, restored 16 / 0. Six new tests, including one end-to-end against a real `EditorView` that asserts the premise (`transactionsSince === 0`, `view.state === stateAtEscape`) before asserting Tab indents.

### Verification

`editor` 669 pass / 1 skip / 0 fail (up 11) · `chat` 777 pass / 0 fail · both typecheck clean · repo lint 10/10, 0 errors.

PR #1289, opened 2026-08-14
fix(editor): use live Milkdown selection updates

### LINKED ISSUE #1288: ReviewEditor: currentSelection is always one transaction stale, so createThread silently anchors the wrong text (or returns null)
> [!IMPORTANT] This body is superseded — read both comments below before acting on anything here
> The mechanism and requested fix below were **refuted in review and are wrong**. The defect is
> real, but it is not about native versus programmatic selections, and the fix is not in
> `createThread`.
>
> Read, in order:
> 1. [The correction](https://github.com/stevekinney/cinder/issues/1288#issuecomment-5289815464) —
>    the actual cause (`currentSelection` lags by exactly one selection-changing transaction,
>    because `notifySelectionChange` reads `view.state` from inside `EditorState.apply`), and the
>    real fix (one line in `notifySelectionChange`, not in `createThread`).
> 2. [The amendment](https://github.com/stevekinney/cinder/issues/1288#issuecomment-5292548733) —
>    two corrections to the correction: "permanently" is too strong (a document change resyncs
>    `currentSelection` through a separate ~200ms-debounced listener), and there is a **worse**
>    case inside that resync window — `createThread` can anchor the OLD selection while the live
>    caret has moved somewhere else entirely, independently reproduced live.
>
> A reader who stops at comment 1 has the mechanism right but is missing the worst-case severity.
>
> The original text is kept below unedited, as the record of what was first reported.

---

## Summary

`ReviewEditor.createThread(body, authorId)` returns `null` for any selection that was not produced by a native user gesture. It reads `currentSelection`, which is populated only through the inner `MarkdownEditor`'s `onselectionchange` (native `selectionchange`), so a consumer following the documented `bind:this` guidance — set a selection via the public `getView()`, then call `createThread` — gets a silent `null` and no thread.

`createDocumentThread` and `createBlockThread` are unaffected and succeed from the identical code path, because neither reads `currentSelection`.

## Reproduction

Verified in a real browser (Chromium, Playwright) against `@lostgradient/editor@0.9.0`, on a bound `ReviewEditor` in `edit` mode.

```ts
// A: programmatic selection through the public imperative surface
const view = editor.getView();
view.focus();
view.dispatch(view.state.tr.setSelection(TextSelection.create(view.state.doc, 44, 53)));
editor.createThread('note', 'author-1');   // -> null, no thread, no onthreadcreate

// B: a DOM Range over the same characters, which is what a drag produces
//    (range set via window.getSelection(), then view.focus())
editor.createThread('note', 'author-1');   // -> null

// C: a real mouse drag across the same paragraph
editor.createThread('note', 'author-1');   // -> 'a9c2084f-…', onthreadcreate fires
```

In A and B, `editor.getSelection()` reports `{from: 44, to: 53, isCollapsed: false}` — a valid non-collapsed selection — at the moment `createThread` returns `null`. That readout comes from `getSelection()`, which resolves through `editorRef.getSelection()` (view-derived), while `createThread` guards on `currentSelection`. The two disagree, and only the latter gates thread creation.

## Why this is a defect rather than a usage error

`createThread`'s only guard-fail signal is `devWarn` plus a `null` return, so in production the call is indistinguishable from a no-op.

More to the point, **the component already contains the fix and applies it elsewhere.** The popover creation path deliberately bypasses `currentSelection` for this exact reason (`review-editor-impl.svelte`, ~line 824):

```
// Get selection directly from the view since currentSelection state
// may not be updated yet when the native selectionchange fires
const view = editorRef?.getView();
if (view) {
  const { from, to } = view.state.selection;
```

So the internal UI path is hardened against `currentSelection` staleness and the public imperative path is not — which inverts the reliability you would expect, given that `bind:this` + `createThread` is what consumers are told to use for anything past the props.

## Requested fix

Have `createThread` resolve its range the same way the popover path does: prefer the live view selection (`view.state.selection`), falling back to `currentSelection`. Keep the collapsed-selection guard, evaluated against whichever source is used.

If the current behavior is intentional — i.e. `createThread` is only ever meant to follow a user gesture — then it should say so in the type docs and README, because nothing at the call site suggests it, and the `null` return is silent.

## Impact

Blocks driving thread creation from tests, scripts, or any restore/seed path that does not involve a human mouse. Found while building `/exercises/review-imperative` in `chatroom` to cover the ReviewEditor imperative surface.

stevekinney:
## Summary
- use Milkdown's live selection in selectionUpdated callbacks
- preserve post-transaction document resync behavior
- add regressions for selection-only staleness and the document debounce window

Closes #1288

### DIFF
--- a/.changeset/bright-selections-1288.md
+++ b/.changeset/bright-selections-1288.md
@@ -1,0 +1,5 @@
+---
+'@lostgradient/editor': patch
+---
+
+Use Milkdown's live selection payload for selection-change notifications so review anchors and toolbar state do not lag one transaction behind.
--- a/packages/editor/src/lib/editor/editor.selection.test.ts
+++ b/packages/editor/src/lib/editor/editor.selection.test.ts
@@ -1,0 +1,70 @@
+/// <reference lib="dom" />
+import { TextSelection } from '@milkdown/prose/state';
+import { afterEach, describe, expect, test } from 'bun:test';
+import { setupHappyDom } from '../test/happy-dom.ts';
+import { createEditor } from './editor.ts';
+import type { EditorState } from './types.ts';
+
+setupHappyDom();
+
+let editorState: EditorState | undefined;
+
+afterEach(async () => {
+  if (editorState) {
+    editorState.markDestroyed();
+    await editorState.editor.destroy();
+    editorState = undefined;
+  }
+});
+
+describe('createEditor selection notifications', () => {
+  test('emits Milkdown selection during a selection-only transaction', async () => {
+    const container = document.createElement('div');
+    document.body.append(container);
+    const selections: Array<{ from: number; to: number; isCollapsed: boolean }> = [];
+
+    editorState = await createEditor(container, {
+      initialContent: 'selection regression',
+      onselectionchange: (selection) => {
+        if (selection) selections.push(selection);
+      },
+    });
+
+    const transaction = editorState.view.state.tr.setSelection(
+      TextSelection.create(editorState.view.state.doc, 1, 10),
+    );
+    editorState.view.dispatch(transaction);
+
+    expect(selections.at(-1)).toEqual({ from: 1, to: 10, isCollapsed: false });
+    container.remove();
+  });
+
+  test('document updates resynchronize selection after the listener debounce', async () => {
+    const container = document.createElement('div');
+    document.body.append(container);
+    const selections: Array<{ from: number; to: number; isCollapsed: boolean }> = [];
+
+    editorState = await createEditor(container, {
+      initialContent: 'document regression',
+      onselectionchange: (selection) => {
+        if (selection) selections.push(selection);
+      },
+    });
+
+    editorState.view.dispatch(
+      editorState.view.state.tr.setSelection(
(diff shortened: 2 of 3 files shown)


stevekinney:
Regression coverage confirms the pre-fix behavior: running the new tests against origin/main failed both selection-only and mapped-selection assertions. Current head includes the requested @lostgradient/editor patch changeset.

stevekinney:
@codex review

stevekinney:
@copilot review

stevekinney:
Addressed Copilot feedback at exact head eae925912: retained the .js internal imports and removed the real-time debounce wait entirely. The regression asserts the mapped live selection immediately during the listener debounce window, avoiding any timeout/wait-threshold increase. Resolved the actionable review thread.

stevekinney:
@codex review

stevekinney:
@copilot review

stevekinney:
@codex review

stevekinney:
@copilot review

stevekinney:
@codex review

stevekinney:
@copilot review

stevekinney:
@codex review

PR #1296, opened 2026-08-14
fix(editor,chat): three accessibility defects found by a three-engine matrix

### LINKED ISSUE #1295: CommentComposer: the inline submit button is unclickable in WebKit — its :focus-within gate revokes pointer-events mid-click
## What happens

In WebKit, clicking the comment composer's **Comment** button does nothing. No `submit`, no `commentcreate` / `threadcreate`, no announcement, composer still populated. The only working path is the undiscoverable Cmd+Enter shortcut.

Chromium and Firefox are unaffected. Found by adding WebKit and Firefox to a consumer's Playwright matrix (`stevekinney/chatroom`), against `@lostgradient/editor@0.9.1`.

## Mechanism

`dist/components/review-editor/comment-composer.svelte` hides the inline submit until the composer has focus:

```css
.comment-composer-inline-submit { position: absolute; … opacity: 0; pointer-events: none; }
.comment-composer-textarea-container:focus-within .comment-composer-inline-submit { opacity: 1; pointer-events: auto; }
```

WebKit does not focus a `<button>` on mousedown. So mid-gesture the textarea blurs, `:focus-within` drops off the container, and the submit flips to `pointer-events: none` **between mousedown and mouseup**. `mouseup` then hit-tests to the textarea, and per spec `click` retargets to the nearest common inclusive ancestor — the wrapper `<div>`. The button never receives `click`, so the form never submits.

Capture-phase trace of one click on the popover composer, WebKit:

```
INITIAL      active=TEXTAREA#…composer            opacity=0.40 pe=auto  focusWithin=true
pointerdown  target=BUTTON.cinder-button          pe=auto  focusWithin=true
mousedown    target=BUTTON.cinder-button          pe=auto  focusWithin=true
focusout     target=TEXTAREA…  active=BODY        pe=none  focusWithin=false
mouseup      target=TEXTAREA…
click        target=DIV.comment-composer-textarea-container    ← never the BUTTON, no submit
```

Chromium and Firefox, same click:

```
mousedown  target=BUTTON.cinder-button   pe=auto  focusWithin=true
focusin    target=BUTTON.cinder-button   active=BUTTON
mouseup    target=BUTTON.cinder-button
click      target=BUTTON.cinder-button
submit     target=FORM.comment-composer
```

## The engine behavior alone is NOT the bug — this is the part worth checking first

A minimal standalone page with **no Svelte and no component code**, comparing a plain textarea+submit form against the same `:focus-within` gate:

```
[chromium] plain: click:b1, submit:plain      gated: click:b2, submit:gated
[firefox]  plain: click:b1, submit:plain      gated: click:b2, submit:gated
[webkit]   plain: click:b1, submit:plain      gated: mousedown:b2, then doc-click target=DIV  ← no click, no submit
```

WebKit submits the **ungated** form exactly like the other two engines. Click-doesn't-focus is benign on its own; it is the stylesheet that converts it into a dead primary affordance. That is why this is a component defect rather than "that's just Safari".

## Controls, all in WebKit

- "Delete thread" — a popover button **not** gated on `:focus-within` — clicks and fires `threaddelete` normally.
- Injecting `.comment-composer-inline-submit { pointer-events: auto !important; }` → `commentcreate` fires and the consumer's assertion passes.
- Adding `preventDefault()` on the button's `mousedown` → `commentcreate` fires.
- Cmd+Enter from the textarea → works. Programmatic `button.click()` → works.

So the submit logic is fine; only the pointer path is broken.

## Requested fix

Call `preventDefault()` on the submit button's `mousedown` so the textarea keeps focus through the whole gesture and `:focus-within` never drops. That is the standard fix for this pattern, and it also addresses the related smell that the primary submit is `opacity: 0` until focus arrives.

Please **don't** fix it by keeping `pointer-events: auto` while `opacity: 0` — that leaves an invisible click target sitting over the textarea.

`CommentComposer` is shared, so one fix should cover the inline composer, the thread popover, and the selection popover.

## Scope

Two consumer tests fail on this in WebKit, both in `review-ssr-and-a11y.e2e.ts` — "adding a document comment and clearing all comments are both announced" (`:627`) and "replying through the popover announces the new comment" (`:985`). Two more (`review-comment-creation.e2e.ts:232`/`:254`) plausibly share it.

**Stated precisely so it can be checked rather than assumed:** all of the above is Playwright's bundled WebKit (a real engine driven with trusted protocol-level input, not a DOM shim), on macOS with Full Keyboard Access off. I have not driven real Safari. The minimal repro above should settle that in under a minute if it matters — and if Safari submits normally, this is mine to retract.

stevekinney:
Closes #1295.

In WebKit, clicking the comment composer's **Comment** button did nothing. No `submit`, no `commentcreate`/`threadcreate`, no announcement, composer still populated. Only the undiscoverable Cmd+Enter worked. Chromium and Firefox were unaffected.

Found from the consumer side in `stevekinney/chatroom`, by adding WebKit and Firefox to its Playwright matrix.

## Mechanism

The inline submit is revealed by `:focus-within` on its container, which also flips it from `pointer-events: none` to `auto`. WebKit does not focus a `<button>` on mousedown, so mid-gesture the textarea blurs, `:focus-within` drops, and `pointer-events` returns to `none` **between mousedown and mouseup**. The mouseup then hit-tests to the textarea, and per spec `click` retargets to the nearest common inclusive ancestor — the wrapper `<div>`. The button never receives `click`; the form never submits.

Capture-phase trace of one click, WebKit vs the others:

```
webkit    mousedown BUTTON  -> focusout TEXTAREA (active=BODY, pe=none) -> mouseup TEXTAREA -> click DIV      (no submit)
chromium  mousedown BUTTON  -> focusin  BUTTON                          -> mouseup BUTTON   -> click BUTTON -> submit FORM
firefox   ditto
```

## Why this is ours and not "just Safari"

A minimal page with **no Svelte and no component code**, comparing a plain textarea+submit form against the same `:focus-within` gate:

```
[chromium] plain: click,submit    gated: click,submit
[firefox]  plain: click,submit    gated: click,submit
[webkit]   plain: click,submit    gated: mousedown only, then click retargets to DIV - no submit
```

WebKit submits the **ungated** form exactly like the other two. Click-doesn't-focus is benign on its own; our stylesheet is what converts it into a dead primary affordance.

Four more controls, all in WebKit, all pointing at the same line:

- "Delete thread" - a popover button *not* gated on `:focus-within` - clicks and fires normally.
- Injecting `pointer-events: auto !important` on the submit -> `commentcreate` fires.
- Adding `mousedown` preventDefault -> `commentcreate` fires.
- Cmd+Enter and programmatic `button.click()` both work, so the submit logic itself was never broken.

## The fix

`preventDefault()` on the submit's `mousedown`, so the textarea keeps focus through the gesture and `:focus-within` never drops.

Deliberately **not** "keep `pointer-events: auto` while `opacity: 0`" - that leaves an invisible click target sitting over the textarea.

`CommentComposer` is shared, so this covers the inline composer, the thread popover, and the selection popover.

## What the test does and does not claim

happy-dom models neither hit-testing, nor `:focus-within`, nor computed `pointer-events`, nor WebKit's focus policy - so it cannot reproduce this failure, and the new test says so in its own header rather than implying coverage it does not have. What it pins is the mechanism the fix rests on: mousedown's default is suppressed, and click's is not. Removing the handler turns the first red; the second is a direction guard, so a fix that overshot into cancelling activation (inert in *every* engine) would be caught.

The end-to-end half is pinned where it is observable - the consumer's WebKit suite, which is what found this.

## Verification

`editor` 681 pass / 0 fail, typecheck 0 errors, `components:check` clean. `lint` is clean once `@lostgradient/markdown` is built - the editor package's `lint` script, unlike the components package's, does not build it first, which is a pre-existing sharp edge unrelated to this change.

### DIFF
--- a/.changeset/artifact-panel-returns-focus.md
+++ b/.changeset/artifact-panel-returns-focus.md
@@ -1,0 +1,11 @@
+---
+'@lostgradient/chat': patch
+---
+
+Return focus when the artifact panel closes, instead of dropping it on `<body>`.
+
+`ArtifactPanel` focuses its Close button on mount so a keyboard user lands inside the panel — deliberate and good — but nothing restored focus on unmount. Closing therefore left `document.activeElement` as `<body>`: the next Tab restarts at the top of the document, and a screen reader announces nothing. Reproduced identically in Chromium, Firefox, and WebKit, so this was never an engine quirk.
+
+The attachment now captures the previously focused element before taking focus and restores it on teardown, guarded on `isConnected` — restoring to a detached node is a silent no-op that would leave the bug in place with no signal. Stated plainly because it is a real limit: a consumer whose close also removes the control that opened the panel still has to manage focus itself, since by teardown the panel has no surviving element of its own to offer either.
+
+Fixes #1299.
--- a/.changeset/snapshot-mode-reaches-editor-content.md
+++ b/.changeset/snapshot-mode-reaches-editor-content.md
@@ -1,0 +1,11 @@
+---
+'@lostgradient/editor': patch
+---
+
+Make `snapshotMode` actually suppress the selection in the editor content, which it never did in any engine.
+
+The rule was authored as `[data-snapshot-mode] *`, and inside a Svelte `<style>` a bare `*` compiles to `:where(.svelte-…)` — so the descendant half could only match elements the component itself rendered. `.milkdown` and `.ProseMirror` are created at runtime by Milkdown with no scope class, so it never applied to them. Chromium looked correct only because Blink inherits `user-select`, which css-ui-4 defines as non-inherited and Gecko implements as such; Firefox reporting `auto` there is the spec-correct value, and is what surfaced this.
+
+`:global(*)` alone is not the whole fix. A real drag inside a snapshot-mode editor selected and repainted in **both** Chromium and Firefox even where `user-select` computed to `none`, because ProseMirror's contenteditable stays selectable regardless. A transparent `::selection` is what makes the surface pixel-stable, which is what the prop documents.
+
+Fixes #1298.
--- a/.changeset/webkit-composer-submit-is-clickable.md
+++ b/.changeset/webkit-composer-submit-is-clickable.md
@@ -1,0 +1,13 @@
+---
+'@lostgradient/editor': patch
+---
+
+Make the comment composer's inline submit button clickable in WebKit, where it did nothing at all.
+
+The button is revealed by `:focus-within` on its container, which also flips it from `pointer-events: none` to `auto`. WebKit does not focus a `<button>` on mousedown — so mid-gesture the textarea blurred, `:focus-within` dropped, `pointer-events` returned to `none` before mouseup, and the mouseup hit-tested to the textarea instead. Per spec the `click` then retargeted to the nearest common ancestor, the wrapper `<div>`, so the button never saw a click and the form never submitted. In Safari the composer's primary affordance was dead and the only working path was the undiscoverable Cmd+Enter.
+
+Worth stating because it decides where the fix belongs: the engine behavior is benign on its own. A minimal page with no component code submits an ungated textarea+button form identically in all three engines; only the `:focus-within` gate breaks it, and only in WebKit. The stylesheet is what turned a spec-legal engine behavior into a broken control.
+
+Fixed by suppressing mousedown's default focus change on the submit, so the textarea keeps focus for the whole gesture and `:focus-within` never drops. Not fixed by keeping `pointer-events: auto` while `opacity: 0`, which would leave an invisible click target over the textarea. `CommentComposer` is shared, so this covers the inline composer, the thread popover, and the selection popover.
+
+Fixes #1295.
--- a/packages/chat/src/lib/components/chat/artifact/artifact-panel.focus-restore.test.ts
+++ b/packages/chat/src/lib/components/chat/artifact/artifact-panel.focus-restore.test.ts
@@ -1,0 +1,69 @@
+/// <reference lib="dom" />
+import { afterEach, describe, expect, test } from 'bun:test';
+import { createRawSnippet, tick } from 'svelte';
+
+import { setupHappyDom } from '../../../test/happy-dom.ts';
+
+setupHappyDom();
+
+const { render } = await import('@testing-library/svelte');
+const { default: ArtifactPanel } = await import('./artifact-panel.svelte');
+
+/**
+ * The panel focuses its Close button on mount so a keyboard user lands inside it.
+ * It gave nothing back on unmount, so closing dropped focus on `<body>` — next
+ * Tab restarts at the top of the document, and a screen reader is silent.
+ * Reproduced identically in Chromium, Firefox, and WebKit before the fix, so it
+ * was never an engine quirk.
+ */
+/** The panel requires a `children` snippet; its content is irrelevant here. */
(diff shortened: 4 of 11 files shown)


stevekinney:
**Scope grew — now closes #1295, #1298, and #1299.**

Batched into this PR rather than cut as three release cycles, per the same-repo batching rule: none of the three had released yet.

**#1298 — `snapshotMode` never reached the editor content, in any engine.** The rule was authored as `[data-snapshot-mode] *`, and inside a Svelte `<style>` a bare `*` compiles to `:where(.svelte-…)` — so the descendant half could only match elements the component itself rendered, never the runtime-created `.milkdown` / `.ProseMirror` subtree. Chromium looked correct only because Blink inherits `user-select`, which css-ui-4 defines as non-inherited; Gecko implements it as such, which is why Firefox surfaced it.

`:global(*)` alone is not the fix. A real drag selected **and repainted** in both Chromium and Firefox even where `user-select` computed to `none`, because ProseMirror's contenteditable stays selectable regardless. A transparent `::selection` is what actually delivers the documented pixel-stable behavior.

**#1299 — `ArtifactPanel` took focus and never gave it back.** It focuses its Close button on mount, which is right; it restored nothing on unmount, so closing left focus on `<body>` — next Tab from the top of the document, silence for a screen reader. Reproduced identically in Chromium, Firefox, and WebKit.

## Test integrity

| test | break | result |
| --- | --- | --- |
| descendant rule escapes scoping | restore the bare `*` | 1 fail |
| panel returns focus to its opener | drop the capture/restore | 1 fail |

The `snapshotMode` test asserts the **authored CSS** rather than computed style, deliberately: the defect *is* the scoping, and happy-dom models neither `::selection` nor scoped-style compilation, so a computed-style assertion there would be measuring the harness. The behavioral half is pinned in the consumer's three-engine suite, which is where it is observable and where this was found. The test says so in its own header.

One limit stated rather than hidden: a consumer whose close also removes the control that opened the panel still gets no restore, because by teardown the panel has no surviving element of its own to offer either. That case needs the consumer to manage focus, the same way `createFocusTrap`'s `restoreFallback` exists for it.

## Verification

`editor` 676 pass / 1 skip / 0 fail, `chat` 779 pass / 0 fail, both typecheck 0 errors, both lint 0 errors, `components:check` clean.

stevekinney:
**Correction to my earlier verification note.** I reported the editor suite as `676 pass / 1 skip / 0 fail`. That was true and incomplete: the same run also printed `2 errors`, and CI failed on it.

The errors were not assertions. Svelte 5's `flushSync` effect-teardown trips a happy-dom divergence on unmount, the `DOMException` escapes through a Promise executor, and Bun reports it as an "unhandled error between tests" — a line separate from pass/fail, which is exactly what I missed by reading only the pass/fail lines.

Attributed rather than assumed: a clean `origin/main` worktree runs the editor suite at `679 pass / 0 fail` with **no** errors line, and this branch produced `2 errors`. So this branch introduced them — the first editor test to unmount a rendered component was enough to expose a gap that was already there.

`packages/components` had already solved this and documented the mechanism; the fix ports that patch into the editor package's own helper. It is deliberately duplicated rather than imported, since a test helper reaching across package boundaries is what `check:consumer-boundaries` exists to prevent — a third consumer should extract it to `@lostgradient/testing` rather than add a copy.

Editor is now **685 pass / 1 skip / 0 fail / 0 errors**; chat **779 pass / 0 fail**; both typecheck and lint clean.

PR #1321, opened 2026-08-15
fix(editor): normalize front matter in toolbar diffStats and generateMarkdownSummary

### LINKED ISSUE #1307: ReviewEditor's toolbar diffStats normalizes front matter as Markdown, so a one-line edit is counted as two — the bug #1285 fixed in generateUnifiedDiff but not here
## What happens

The control bar's change counter and the diff panel disagree about how many lines changed, and the toolbar is the one that is wrong. A **one-line front-matter edit** is reported as **2 modified lines** by the toolbar and **0** by the panel's body diff.

This is the same defect class [#1285](https://github.com/stevekinney/cinder/pull/1285) fixed in `generateUnifiedDiff` — its own comment says so — but the toolbar's `diffStats` never got the fix.

## Repro

`@lostgradient/editor@0.9.2`, `@lostgradient/markdown` pipeline, editing only `owner: jane` → `owner: bob`:

```js
const orig = '---\ntitle: Release Plan\nowner: jane\n---\n\n# Release Plan\n\nShip it.\n';
const cur  = '---\ntitle: Release Plan\nowner: bob\n---\n\n# Release Plan\n\nShip it.\n';

normalize(orig) === "---\n\ntitle: Release Plan\nowner: jane\n-----------\n\n# Release Plan\n\nShip it.\n"
normalize(cur)  === "---\n\ntitle: Release Plan\nowner: bob\n----------\n\n# Release Plan\n\nShip it.\n"

getDiffStats(computeLineDiff(normalize(orig), normalize(cur)))  // {added:0, removed:0, modified:2}
// panel, body-only:                                            // {added:0, removed:0, modified:0}
```

## Mechanism

`normalize()` is a Markdown pipeline with **no front-matter step**. Handed a whole document it reads `---\ntitle: …\n---` as a thematic break followed by a **setext heading**, and re-serializes the underline at the content's width. So `jane` → `bob` shortens the line by three characters and the underline goes from 11 dashes to 10 — a second changed line that exists only as an artifact of normalization.

- Toolbar: `dist/components/review-editor/review-editor-impl.svelte:189-198` — `diffStats` calls `normalize(original)` / `normalize(value)` on the WHOLE document.
- Panel: `dist/components/diff-viewer/diff-viewer.svelte:79-100` splits front matter off first and diffs the body.
- Already-fixed sibling: `dist/export/unified-diff.js:160-192`, whose comment reads *"`normalize()` is a Markdown pipeline with no front-matter step… sidesteps this by normalizing only the body; the export path must do the same."*

That last comment is the fix this file needs and did not get.

## Requested fix

Have `diffStats` use the same front-matter-aware path the export already uses — split front matter off, normalize the body only, and account for front-matter changes separately — so the toolbar's count matches the panel's and matches the edit the user made.

## Related, and worth deciding together

The two normalizers disagree on ordinary Markdown too, not just front matter. Each of these is reported identical by the pipeline normalizer and different by the panel's regex one:

| input pair | pipeline | regex |
| --- | --- | --- |
| setext vs ATX heading | equal | differ |
| `_emphasis_` vs `*emphasis*` | equal | differ |
| `1. 1. 1.` vs `1. 2. 3.` | equal | differ |
| `***` vs `---` | equal | differ |

There are in fact **five** normalization paths in the package, not two — `normalize` (toolbar), `normalizeForDiff` (panel), `normalizeDocument` (unified-diff export, front-matter-aware), `normalizeForEditor` (MarkdownEditor, byte-identical regex chain to the panel's), and `normalizeSafely`, which picks between the pipeline and regex normalizers **depending on whether a lazy dynamic import has resolved yet** — so the same input can normalize two different ways in one session.

Filing the front-matter count as the actionable defect; the wider set is context for whoever picks it up, since a fix that only patches `diffStats` leaves four paths still disagreeing.

Found via `stevekinney/chatroom` ROADMAP item `X-2`, carried over from #1285's own "not included" list.

### LINKED ISSUE #1318: generateMarkdownSummary has no normalization step, so it disagrees with generateUnifiedDiff about whether a blank-line-only edit is a change
## What happens

`generateUnifiedDiff` and `generateMarkdownSummary` can report opposite verdicts on the exact same `ReviewState`: one says nothing changed, the other says there was a real edit.

Root cause: `generateMarkdownSummary` has no normalization step at all — it runs `computeLineDiff` directly on the raw `original`/`current` strings. `generateUnifiedDiff` (default `normalizeInputs: true`) runs both strings through `normalizeDocument`, which strips leading blank lines from the body before re-serializing. So any input where the normalized forms collapse to the same document but the raw strings differ makes the two functions disagree about whether an edit happened.

## Repro 1 — front matter (matches this package's own `generateUnifiedDiff` docs, `@lostgradient/editor@0.10.0`)

Same front matter, same body text. The only difference between `original` and `current` is how many blank lines separate the closing `---` from the body (one vs. three):

```js
import { generateUnifiedDiff, generateMarkdownSummary } from '@lostgradient/editor/export';

const fm = ['---', 'title: Release Plan', 'draft: true', '---'].join('\n');
const original = `${fm}\n\nAlpha line.`;
const current = `${fm}\n\n\n\nAlpha line.`;

const state = { original, content: current, threads: [] };

generateUnifiedDiff(state).diff
// => "" (hunks: 0 — reports NO change)

generateMarkdownSummary(state).markdown
// => "## Changes Made\n\nThe following edits were made to the document:\n\n### Lines 4-6\n\n```diff\n ---\n \n+\n+\n Alpha line.\n```\n"
// (changeCount: 2 — reports a real edit)
```

## Repro 2 — no front matter at all (proves the root cause, not a front-matter-parsing gap)

```js
const original = 'Alpha.\n\nBeta.';
const current = 'Alpha.\n\n\n\nBeta.';
const state = { original, content: current, threads: [] };

generateUnifiedDiff(state).diff        // => "" (hunks: 0)
generateMarkdownSummary(state).markdown
// => "## Changes Made\n\nThe following edits were made to the document:\n\n### Lines 1-3\n\n```diff\n Alpha.\n \n+\n+\n Beta.\n```\n"
```

`normalize()`'s own blank-line collapsing (`.replace(/\n{3,}/g, '\n\n')` plus the leading-newline strip) is enough on its own — front matter just happens to be where this route's fixtures exercise it, because `normalizeDocument`'s front-matter separator is *always* exactly one hardcoded `\n` regardless of how many blank lines separated the original front matter from the body.

## A second symptom of the same missing step: CRLF

`generateUnifiedDiff`'s `normalizeDocument` also does a universal `content.replace(/\r\n?/g, '\n')` before anything else. `computeLineDiff` does no such thing — it splits only on `\n` (`packages/markdown/src/diff/line-diff.ts:137`, `splitLines`), so a CRLF document's lines keep a literal trailing `\r` in their `text`. That `\r` lands verbatim inside `generateMarkdownSummary`'s ` ```diff ` code fence, while the same document through `generateUnifiedDiff` comes out CRLF-normalized. Not filing a separate repro for this since it's the same missing normalization step, surfacing a different way.

## Source (as of `origin/main` @ `65025eb58c59b0beca93acc089c3424f55f3224a`, matches published `@lostgradient/editor@0.10.0`)

- `packages/editor/src/lib/export/unified-diff.ts:249` — `normalizeDocument`, called at `unified-diff.ts:154` and `:157` for `original`/`current` whenever `normalizeInputs` is true (the default). `splitDocument` (`unified-diff.ts:228`) does the CRLF normalization and front-matter split.
- `packages/editor/src/lib/export/markdown-summary.ts:50-53` — `generateMarkdownSummary` compares `state.original ?? ''` against `state.content` directly (`original !== current`), then `generateChangesSection` (`markdown-summary.ts:95`) calls `computeLineDiff(original, current)` at `markdown-summary.ts:100` with no normalization of either string.
- `packages/markdown/src/diff/line-diff.ts:82` — `computeLineDiff` itself: no CRLF handling, no blank-line collapsing, no front-matter awareness.

## Reachability

No UI keystroke sequence can type this directly — ProseMirror collapses adjacent blank paragraphs in its own document model. But `ReviewState.original` is consumer-supplied verbatim (it's never routed through the editor), and `ReviewEditor.setState` round-trips persisted content the same way. Any external tool, hand-edited save file, or upstream system that changes blank-line spacing near front matter — or CRLF vs LF line endings anywhere — produces exactly this state pair through the public surface.

## Same defect class as #1307 and #1285

#1285 fixed this exact front-matter-normalization gap in `generateUnifiedDiff` itself (that's where `normalizeDocument`/`splitDocument` came from). #1307 found the same class in the toolbar's `diffStats`, which still calls the front-matter-blind `normalize()` directly. This issue is the third instance: `generateMarkdownSummary` has never had *any* normalization step, front-matter-aware or otherwise — it's not that it forgot the front-matter case, it's missing the whole thing `generateUnifiedDiff` runs by default.

## Requested fix

Run `generateMarkdownSummary`'s `original`/`current` through the same `normalizeDocument` (or an exported equivalent) that `generateUnifiedDiff` uses by default, before diffing — with an options flag mirroring `normalizeInputs` if callers need the raw comparison too, matching `generateUnifiedDiff`'s own API shape.

Found via `stevekinney/chatroom` ROADMAP item `X-2`, whose own test suite (`review-front-matter.e2e.ts`) already documented that `generateUnifiedDiff` and `generateMarkdownSummary` normalize differently and pinned it as a tripwire; this is the disagreeing input that tripwire was watching for.

stevekinney:
## Summary

Same defect class #1285 already fixed in `generateUnifiedDiff`, spread to two consumers that
never got the fix:

- **#1307** — ReviewEditor's toolbar `diffStats` normalized the whole document (front matter and
  body together) with `normalize()`, a Markdown pipeline with no front-matter step. Handed
  `---\ntitle: …\n---`, it reads the fences as a thematic break plus a setext heading and
  re-emits the closing underline at the new content width, so a one-line front-matter value edit
  (`owner: jane` → `owner: bob`) counted as **two** modified lines instead of one.
- **#1318** — `generateMarkdownSummary` ran `computeLineDiff` directly on raw strings with *zero*
  normalization (no CRLF handling, no front-matter awareness, no blank-line collapsing), while
  `generateUnifiedDiff` normalizes both inputs by default. Two documents differing only in
  blank-line padding after front matter (or only in line-ending style) made the two exports
  disagree about whether an edit happened at all — one said "nothing changed", the other emitted
  a two-line diff.

Both are fixed by extending the exact front-matter-aware normalization `generateUnifiedDiff`
already used, factored out of `unified-diff.ts` into a new `export/normalize-document.ts` so
`generateUnifiedDiff`, the toolbar's new `computeReviewEditorDiffStats`, and
`generateMarkdownSummary` (now via a `normalizeInputs` option defaulting to `true`, mirroring
`generateUnifiedDiff`'s own) all share one implementation instead of each carrying its own copy.
`normalizeDocument`/`splitDocument` are exported from `@lostgradient/editor/export` so a future
consumer reuses them too.

`computeReviewEditorDiffStats` is pulled out of `review-editor-impl.svelte` into its own module,
since the component itself is too heavy to mount in happy-dom (Milkdown needs a real browser DOM
— see `review-editor.snapshot-mode.test.ts`'s existing note on this). A source pin in
`review-editor-diff-stats.test.ts` covers the wiring the behavioral test can't reach: that the
component actually calls the fixed helper rather than reintroducing an inline `normalize()` call.

## Revert verification (every fix has a test proven to fail without it)

- Reverting `review-editor-diff-stats.ts`'s body back to a bare `normalize()` call reproduces
  #1307's exact repro: `{ added: 0, removed: 0, modified: 2 }` instead of `modified: 1`.
- Reverting the `diffStats` wiring in `review-editor-impl.svelte` back to an inline
  `normalize()`-based computation fails the source-pin test (proves the wiring, not just the
  helper, is covered).
- Reverting `markdown-summary.ts`/`types.ts` to `origin/main` fails 5 of the 6 new tests in
  `markdown-summary.test.ts`'s `normalization (cinder#1318)` block — the front-matter repro, the
  no-front-matter repro, both CRLF cases, and the with-feedback variant. The 6th test
  (`normalizeInputs: false restores the raw comparison`) intentionally passes against both old and
  new code — it pins the opt-out contract, not the bug itself, since the old code's behavior *was*
  the unnormalized comparison.

All three reverts were restored and re-verified green afterward, confirmed by content hash
against a backup, not just by eye.

## Scope

The wider 5-normalizer divergence #1307 documents in its "Related, and worth deciding together"
table (`normalize` vs `normalizeForDiff` vs `normalizeForEditor` vs `normalizeSafely`, plus the
`normalizeSafely` dynamic-import race) is **deliberately untouched** — the issue frames it as
context for whoever picks it up next, not part of this fix. `diff-viewer.svelte`'s own
already-correct front-matter splitting is untouched for the same reason.

## Pre-existing, unrelated: editor's local coverage ratchet

`bun run test:coverage` for `@lostgradient/editor` currently fails locally on the Svelte
function-coverage ratchet (`78.79% < 81.22%`) — reproduced identically on unmodified
`origin/main` in a fresh worktree before any of this PR's changes, so it is not something this PR
introduces or should paper over. It also isn't part of the CI gate for this package: both
`unit-tests.yaml` and `main-green.yaml` run `turbo run test --filter=@lostgradient/editor` (plain,
uncoveraged) for editor; only `@lostgradient/chat` gets `test:coverage` treatment in either
workflow. `turbo run test`, `typecheck`, and `lint` (the actual CI-equivalent gates) are all green
on this branch — see test plan.

## Test plan

- [x] `bunx turbo run test --filter=@lostgradient/editor` — 709 pass, 0 fail, 1 skip (pre-existing,
      unrelated)
- [x] `bunx turbo run typecheck --filter=@lostgradient/editor` — 0 errors, 0 warnings
- [x] `bunx turbo run lint --filter=@lostgradient/editor` — 0 warnings, 0 errors
- [x] `bun run --filter=@lostgradient/editor components:check` — OK
- [x] `bun run --filter=@lostgradient/editor platform:audit -- --strict` — OK
- [x] `bun run --filter=@lostgradient/editor colors:audit -- --strict` — OK
- [x] `bun run --filter=@lostgradient/editor tokens:audit -- --strict` — OK
- [x] `bunx prettier --check` on every touched file — OK
- [x] `bunx changeset status` — recognizes both changesets, `@lostgradient/editor` patch
- [x] Three separate revert-and-restore cycles, each producing a red test, each restore verified
      by hash

Fixes #1307
Fixes #1318

### DIFF
--- a/.changeset/markdown-summary-normalization.md
+++ b/.changeset/markdown-summary-normalization.md
@@ -20,7 +20,10 @@ edit had touched.
 `generateUnifiedDiff`'s own default, and normalizes through the same shared `normalizeDocument`
 (now factored out of `unified-diff.ts` into `export/normalize-document.ts` so `diffStats`, this
 function, and `generateUnifiedDiff` share one implementation — see the cinder#1307 changeset in
-this same batch). Pass `normalizeInputs: false` to restore the previous raw-string comparison.
+this same batch). Pass `normalizeInputs: false` for a byte-for-byte raw comparison, including
+CRLF line endings — stricter than `generateUnifiedDiff`'s own `normalizeInputs: false`, which
+still folds CRLF to LF even with normalization off (a pre-existing quirk of that function, not a
+contract this new option inherits).
 
 This changes `generateMarkdownSummary`'s default output for formatting-only and
 blank-line-only edits: they no longer appear in the "Changes Made" section or count toward
--- a/.changeset/markdown-summary-normalization.md
+++ b/.changeset/markdown-summary-normalization.md
@@ -1,0 +1,28 @@
+---
+'@lostgradient/editor': patch
+---
+
+Fix `generateMarkdownSummary` disagreeing with `generateUnifiedDiff` about whether an edit
+happened (cinder#1318).
+
+`generateMarkdownSummary` ran `computeLineDiff` directly on the raw `original`/`current`
+strings, with no normalization at all — no CRLF handling, no front-matter awareness, no
+blank-line collapsing. `generateUnifiedDiff` normalizes both inputs by default
+(`normalizeInputs: true`) through `normalizeDocument`, which strips leading blank lines from the
+body before re-serializing and reattaches front matter through a single canonical separator. So
+two documents whose front matter and body were byte-identical, differing only in how many blank
+lines separated the closing `---` from the body (or only in line-ending style), made
+`generateUnifiedDiff` report zero hunks ("nothing changed") while `generateMarkdownSummary`
+reported a two-line edit — genuinely disagreeing outputs for a `ReviewState` no consumer-visible
+edit had touched.
+
+`generateMarkdownSummary` now takes a `normalizeInputs` option, defaulting to `true` to match
+`generateUnifiedDiff`'s own default, and normalizes through the same shared `normalizeDocument`
+(now factored out of `unified-diff.ts` into `export/normalize-document.ts` so `diffStats`, this
+function, and `generateUnifiedDiff` share one implementation — see the cinder#1307 changeset in
+this same batch). Pass `normalizeInputs: false` to restore the previous raw-string comparison.
+
+This changes `generateMarkdownSummary`'s default output for formatting-only and
+blank-line-only edits: they no longer appear in the "Changes Made" section or count toward
+`changeCount`, matching what `generateUnifiedDiff` and the diff panel already reported for the
+same input. CRLF line endings no longer leak a literal `\r` into the ` ```diff ` code fence.
--- a/.changeset/review-editor-toolbar-diff-stats-front-matter.md
+++ b/.changeset/review-editor-toolbar-diff-stats-front-matter.md
@@ -1,0 +1,21 @@
+---
+'@lostgradient/editor': patch
+---
+
+Fix ReviewEditor's toolbar change counter over-counting front-matter edits (cinder#1307).
+
+The toolbar's `diffStats` normalized the whole document — front matter and body together —
+with `normalize()`, a Markdown pipeline with no front-matter step. Handed `---\ntitle: …\n---`
+it read the fences as a thematic break plus a setext heading and re-emitted the closing
+underline at the new content's width, so shortening `owner: jane` to `owner: bob` changed the
+value line AND (because it no longer recognized the fence) the underline beneath it — one real
+edit counted as two modified lines, while the diff panel and `exportUnifiedDiff()` (already
+fixed for this in cinder#1285) correctly reported one.
+
+`diffStats` is now computed by `computeReviewEditorDiffStats`, a small function pulled out of
+`review-editor-impl.svelte` into its own module so it is testable without mounting the
+component (which needs a real browser DOM for Milkdown). It calls the same front-matter-aware
+`normalizeDocument` `generateUnifiedDiff` already used, split out of `unified-diff.ts` into
+`export/normalize-document.ts` so it has exactly one implementation instead of one per
+consumer — see the cinder#1318 changeset in this same batch for the second consumer that fix
+reaches.
--- a/.changeset/review-editor-toolbar-diff-stats-front-matter.md
+++ b/.changeset/review-editor-toolbar-diff-stats-front-matter.md
@@ -19,3 +19,8 @@ component (which needs a real browser DOM for Milkdown). It calls the same front
 `export/normalize-document.ts` so it has exactly one implementation instead of one per
 consumer — see the cinder#1318 changeset in this same batch for the second consumer that fix
 reaches.
+
+The same fix also reaches `createReviewEditorState` (`@lostgradient/editor/review-editor`'s
+exported state-manager factory): it had its own copy of the same bare-`normalize()` `diffStats`
+computation, a second public API path that disagreed with the fixed toolbar about the same
+content until it was routed through `computeReviewEditorDiffStats` too.
--- a/packages/editor/src/lib/components/review-editor/review-editor-diff-stats.test.ts
+++ b/packages/editor/src/lib/components/review-editor/review-editor-diff-stats.test.ts
@@ -1,0 +1,113 @@
+/**
+ * Tests for the ReviewEditor toolbar's diff-stats computation (cinder#1307).
+ *
+ * `computeReviewEditorDiffStats` is exercised directly here rather than through
+ * a mounted `ReviewEditor`: the component is too heavy to mount in happy-dom
+ * (Milkdown + ProseMirror internals fail in non-browser DOMs — see
+ * `review-editor.snapshot-mode.test.ts`), but this computation only touches
+ * its two string arguments and needs no DOM at all.
+ *
+ * A source pin below covers the half this file's behavioral test cannot: that
+ * `review-editor-impl.svelte` actually calls this function for `diffStats`
+ * rather than reintroducing a direct `normalize()` call that would make the
+ * bug come back invisibly to this test file.
+ */
+import { describe, expect, test } from 'bun:test';
+import { computeReviewEditorDiffStats } from './review-editor-diff-stats.ts';
+
+describe('computeReviewEditorDiffStats', () => {
+  test('counts a one-line front-matter value edit as one modified line, not two', () => {
+    // cinder#1307's exact repro: normalize() has no front-matter step, so handed
+    // the whole document it re-reads the `---` fences as a thematic break plus a
+    // setext heading and re-emits the closing underline at the new content
+    // width. Shortening `owner: jane` to `owner: bob` changes nothing about the
+    // dash count, but normalize() shortens BOTH the value line and (because it
+    // no longer thinks it is a fence) the underline it invents beneath it — one
+    // real edit surfacing as two modified lines.
+    const original = '---\ntitle: Release Plan\nowner: jane\n---\n\n# Release Plan\n\nShip it.\n';
+    const current = '---\ntitle: Release Plan\nowner: bob\n---\n\n# Release Plan\n\nShip it.\n';
+
+    expect(computeReviewEditorDiffStats(original, current)).toEqual({
+      added: 0,
+      removed: 0,
+      modified: 1,
+    });
+  });
+
+  test('agrees with generateUnifiedDiff on the same front-matter edit', () => {
+    // generateUnifiedDiff already got the cinder#1285 fix. Both counts should
+    // describe the same edit the same way: one line changed (getDiffStats
+    // counts a paired -/+ as one "modified" line; generateUnifiedDiff reports
+    // it as one addition and one deletion).
+    const original = '---\ntitle: Release Plan\nowner: jane\n---\n\nShip it.\n';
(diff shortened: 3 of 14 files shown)


stevekinney:
Addressed all seven review threads in af3c18ff9. Summary, thread by thread:

**Fixed — real gap, thanks for catching it.** `createReviewEditorState` (`@lostgradient/editor/review-editor`'s exported state-manager factory) is a second, independent public path to the same "how many lines changed" question the ReviewEditor toolbar answers, and it still called bare `normalize()` — so a consumer building their own review UI on that API saw the exact #1307 overcount (`modified: 2`) even after the toolbar itself was fixed. Routed it through `computeReviewEditorDiffStats`. New test in `review-editor-state.test.ts`, revert-verified (reproduces `modified: 2` without the fix, restored and confirmed green).

**Fixed — contract mismatch, three duplicate comments about the same thing.** `normalizeInputs: false` was documented as a raw, verbatim comparison "including CRLF" but the implementation still folded CRLF→LF regardless (mirroring `generateUnifiedDiff`'s own always-on CRLF fold). Decided in favor of making the implementation match the promise rather than weakening the docs: `normalizeInputs: false` is now a true byte-for-byte comparison, CRLF included. This is deliberately *stricter* than `generateUnifiedDiff`'s own `normalizeInputs: false` — that always-on CRLF fold is a pre-existing quirk of that function, not a contract this new option needs to inherit, since it's brand new in this PR and I can define its contract correctly from the start. Doc comment and changeset updated to say so explicitly; new revert-verified test (`normalizeInputs: false preserves CRLF differences too`) pins it.

**Investigated, reproduced, refuted as pre-existing/out of scope — line-number drift.** Real: with several blank lines before a genuine edit and `contextLines: 0`, `generateMarkdownSummary`'s `### Lines X-Y` heading reports a position in normalized-document space, not the raw source (confirmed: reports "Lines 3-3" for an edit that's actually on source line 5). But `generateUnifiedDiff` has had the identical property since #1285 — its hunk headers also describe positions in normalized-document space, confirmed by generating a diff for the same input whose `@@` header line count doesn't match the raw original's blank-line count. This isn't a regression #1321 introduces; it's the same trade-off the normalize-then-diff strategy already made, now shared consistently instead of being unique to one function. Fixing it needs a position-mapping layer back to raw source lines, which is real scope beyond what #1307/#1318 asked for. Worth a follow-up issue if it matters to a consumer; not blocking this one.

**Investigated, reproduced, refuted as pre-existing/out of scope — permissive front-matter fence.** Real: `parseFrontMatter` treats any document opening with `---` as front matter up to the next `---`, even when the content between them doesn't parse as YAML (confirmed: a document bracketing an ordinary Markdown list between two `---` lines gets that list preserved byte-for-byte instead of normalized, so a cosmetic `*` → `-` marker change inside it is wrongly reported as an edit). But `splitDocument`/`normalizeDocument` here are byte-identical to what `unified-diff.ts` already shipped (diffed against `origin/main` to confirm) — this PR only *moved* that code into a shared module, it didn't change `parseFrontMatter`'s own fence-detection rules. `generateUnifiedDiff` already had this exact false-positive before this PR, for any document shaped this way. Not something #1321 introduces or worsens; out of scope for #1307/#1318, and any fix belongs in `parseFrontMatter` itself (a different package).

All threads resolved. CI re-running on af3c18ff9.

PR #1330, opened 2026-08-15
fix(markdown,editor): reject non-YAML front matter and map diff line numbers to source

### LINKED ISSUE #1324: editor: normalized-document line numbers in generateMarkdownSummary/generateUnifiedDiff drift from the actual source position
## Summary

Both `generateMarkdownSummary`'s `### Lines X-Y` headings and `generateUnifiedDiff`'s `@@` hunk
headers report line numbers computed against the *normalized* document (after
`normalizeDocument`/`splitDocument` strips leading blank lines from the body and re-attaches
front matter with a single `\n`), not the original source document. When a document contains
extra blank lines before a genuine edit, the reported line number points at the wrong line in
either `state.original` or `state.content`.

Found and refuted-as-out-of-scope during review of
[#1307](https://github.com/stevekinney/cinder/issues/1307)/[#1318](https://github.com/stevekinney/cinder/issues/1318)'s
fix (PR [#1321](https://github.com/stevekinney/cinder/pull/1321)) — it is **pre-existing
behavior inherited unchanged from [#1285](https://github.com/stevekinney/cinder/issues/1285)**,
not something that PR introduces, and fixing it needs a real position-mapping layer back to
source rather than a small normalization fix, so it's out of scope there.

## Repro

```
original = 'Alpha\n\n\n\nOriginal text\n'
current  = 'Alpha\n\n\n\nChanged text\n'
```

With `contextLines: 0`:

- `generateMarkdownSummary` reports `### Lines 3-3` — but the edited line is line 5 in both the
  raw `original` and `current` strings.
- `generateUnifiedDiff` reports a hunk header (`@@ -1,3 +1,3 @@` for this input) that also
  addresses the *normalized* document's line numbering, not the raw source's — confirmed the
  resulting patch cannot be applied via `git apply` to the raw `original`/`current` strings,
  only to their normalized form.

## Suggested fix

Either compute change locations against the original (pre-normalization) document and use the
normalized document only for the equality/diff comparison itself, or maintain an explicit
line-number mapping from normalized positions back to source positions and use it when rendering
`### Lines` headings and `@@` hunk headers. Whichever direction is chosen, both
`generateMarkdownSummary` and `generateUnifiedDiff` should end up consistent with each other,
since they currently share the same underlying drift.

### LINKED ISSUE #1325: markdown: parseFrontMatter treats any `---`-delimited prefix as front matter even when the content isn't valid YAML
## Summary

`parseFrontMatter` (and the `extractFrontMatterSegments` helper it calls) decides a document
has front matter purely from the presence of a `---` opening line and a later `---` closing
line — it never requires the content between them to actually be valid YAML before treating that
span as a front-matter block. When the YAML parse subsequently fails, `hasFrontMatter` is still
`true` and the between-delimiters content is still split off from the body as `raw`/`text`, with
only `data` coming back `null`.

Consumers that treat `hasFrontMatter: true` as "this document has a real front-matter block, and
its front matter and body should be normalized/diffed separately" (e.g.
`@lostgradient/editor`'s `normalizeDocument`/`splitDocument`, used by `generateUnifiedDiff`,
`generateMarkdownSummary`, and `ReviewEditor`'s toolbar `diffStats`) end up preserving a
false-positive "front matter" span byte-for-byte instead of normalizing it as ordinary Markdown
body content.

Found and refuted-as-out-of-scope while reviewing
[stevekinney/cinder#1307](https://github.com/stevekinney/cinder/issues/1307)/[#1318](https://github.com/stevekinney/cinder/issues/1318)'s
fix (PR [#1321](https://github.com/stevekinney/cinder/pull/1321)) — the behavior is unchanged
from what `@lostgradient/editor`'s `splitDocument` already inherited from
[#1285](https://github.com/stevekinney/cinder/issues/1285), and the root cause lives in this
package (`@lostgradient/markdown`), not in editor's normalization logic, so it's out of scope
for that PR.

## Where

`packages/markdown/src/pipeline/frontmatter.ts`, `extractFrontMatterSegments` (matches any line
satisfying `/^---[ \t]*\r?$/m` after the first line as the closing delimiter, with no check on
what's between) and `parseFrontMatter` (catches a YAML parse failure but still returns
`hasFrontMatter: true`).

## Repro

Two documents whose only difference is a `*` vs `-` list marker inside what looks like (but
isn't) front matter:

```
A: '---\n\n* one\n\n---\n\nBody.\n'
B: '---\n\n- one\n\n---\n\nBody.\n'
```

Both fail YAML parsing (`* one` / `- one` alone isn't valid front-matter data), so semantically
these are two Markdown documents that both start with a thematic break, contain a list item, and
end with a second thematic break — `* one` and `- one` are the same Markdown list, just a
different marker character, and any Markdown-aware normalizer treats them as equivalent. But
because `parseFrontMatter` still reports `hasFrontMatter: true` for both, a consumer like
`@lostgradient/editor`'s `splitDocument` preserves that whole span verbatim instead of running it
through Markdown normalization — so the two documents normalize to two *different* strings, and a
diff between them is reported as a real edit rather than the equivalent-formatting no-op it would
be anywhere else in the document.

## Suggested fix

Require the between-delimiters content to actually parse as valid YAML (and plausibly be
object-shaped front-matter data, not e.g. a bare scalar) before setting `hasFrontMatter: true`;
otherwise return `hasFrontMatter: false` and treat the whole document as body, the same as the
"no closing delimiter found" branch already does. This changes `parseFrontMatter`'s public
contract (a document that previously reported `hasFrontMatter: true` with `data: null` on a
parse failure would now report `hasFrontMatter: false`), so downstream consumers and this
package's own tests should be checked for reliance on the current behavior before changing it.

stevekinney:
## Summary

Two related bugs, fixed together because #1325's fix changes what #1324's fix has to be built and tested against.

- **#1325** (`packages/markdown`) — `parseFrontMatter` (via `extractFrontMatterSegments`) classified any `---`-delimited prefix as front matter, with no check on whether the content between the delimiters was actually valid, object-shaped YAML. A YAML parse failure was caught but `hasFrontMatter` still came back `true` with `data: null`. Two consequences: (a) a document that starts with `---`, contains a Markdown list, and closes with a second `---` (`- one` is valid YAML, just a sequence, not a mapping) got misclassified, and (b) genuinely invalid YAML (`* one` — an alias reference with no anchor) did too. Downstream, `@lostgradient/editor`'s `normalizeDocument`/`splitDocument` treat `hasFrontMatter: true` as license to preserve that span byte-for-byte instead of Markdown-normalizing it, so two documents differing only in list-marker style (`* one` vs `- one` — the same list, different marker) normalized to different strings and a diff between them reported a real edit instead of the formatting-only no-op it is everywhere else.

- **#1324** (`packages/editor`) — `generateUnifiedDiff`'s `@@` hunk headers and `generateMarkdownSummary`'s `### Lines X-Y` headings are computed against the *normalized* documents (diffed instead of the raw inputs, deliberately, so formatting-only differences don't get reported as edits), but `normalizeDocument` also changes line *count* (collapsed blank-line runs, a dropped front-matter separator, a folded Setext underline). Both functions reported the normalized-space line index directly, so any header/heading past a point where normalization changed the line count pointed at the wrong line — confirmed with `git apply --check` against the raw `original` string, which rejected the patch outright.

**This PR went through seven review rounds, and the fix mechanism for #1324 changed shape twice as a result — the sections below describe the final, current state, not the first pass. See "Revision history" for what changed and why.**

## Fix mechanism

**#1325**: `parseFrontMatter` now requires the content between the delimiters to parse as YAML *and* be object-shaped (a key/value mapping — not a bare scalar, sequence, or `null`) before reporting `hasFrontMatter: true`. Otherwise it returns `hasFrontMatter: false` and the whole document as `body`, same as the existing "no closing delimiter" branch. An intentionally empty front-matter block (`---\n---\n`) is unaffected — blank content was never a parse failure. `FrontMatterParseResult` also carries a third state, `fencePresent` (`true` whenever a `---`...`---` span exists at all, regardless of validity), so a caller deciding whether it's safe to prepend a *new* fence doesn't misread "not valid front matter" as "no fence to collide with." This package's own `normalizeWithFrontMatter` and `contentEqualsWithFrontMatter` inherit the fix and preserve/compare the parsed `raw` text verbatim whenever `hasFrontMatter` is true but `data` is null (covers both a genuinely blank block and a comment-only one) rather than treating "no data" as "nothing to preserve."

**#1324**: `buildSourceLineMap` (`packages/editor/src/lib/export/source-line-map.ts`) maps normalized-document line numbers back to source-document line numbers. It is **not** a string-comparison heuristic — it parses both `source` and `normalized` (`parseOrThrow`, `@lostgradient/markdown/pipeline`) and walks the two ASTs in lockstep, pairing corresponding nodes by structural *type* (a `heading` is never paired with a `thematicBreak`, no matter what text either serializes to). Each paired leaf node contributes an exact source/normalized line-span anchor, linearly interpolated when the two spans differ in length (e.g. a Setext heading's two source lines folding into its one-line ATX form). `generateUnifiedDiff` remaps hunk `originalStart`/`currentStart` through this map; `generateMarkdownSummary` maps its start *and* end line independently, since a collapsed run inside the displayed range makes `start + count - 1` undercount even once `start` is correctly mapped. `normalizeInputs: false` needs no mapping — the diffed strings already are the (CRLF-folded) source text. Both callers go through `buildSourceLineMapCached`, an LRU cache over `(source, normalized)` pairs.

**Interaction the combined PR exists to prove**: `unified-diff.test.ts` and `markdown-summary.test.ts` both include a test using an invalid-YAML `---`-prefix (now body, post-#1325) plus a blank-line-collapsing edit, asserting the reported line number matches raw source. This only passes if both fixes are built against each other.

## Revision history on the #1324 mechanism (why it isn't what round 1 shipped)

Round 1 aligned normalized and source lines by exact string match, falling back to interpolation for anything that didn't match verbatim. Rounds 2–3 found that too narrow — `normalize()` can rewrite a surviving line (a collapsed heading, a canonicalized list marker) in the same pass that deletes a genuinely-removed neighboring line, and treating the rewritten line as "unmatched" let the deletion get misattributed onto its position — so those rounds added string canonicalization (first a hand-rolled marker regex, then calling the real `normalize()` per line).

**Round 4 proved that entire string-based approach unsound, not just incompletely enumerated.** A Setext heading's underline (`Title\n---`) and a bare thematic break (`***`) are different mdast node types that can both canonicalize to the identical string `---`; no amount of per-line string canonicalization can recover which source line a normalized `---` came from once two structurally different nodes produce the same text. The same round found this machinery runs on *every content edit*, not just a user-triggered export: `ReviewEditor`'s hidden `formDiff`/`formSummary` inputs are `$derived` whenever the component has a `name`, so it re-ran per keystroke, not per export — a performance claim an earlier version of this PR's changeset stated incorrectly (see "Performance" below, corrected).

**The round-4 fix**: rebuild the alignment from the AST instead of reconstructing it after serialization by comparing text. This is what's described in "Fix mechanism" above, and it resolves the Setext/thematic-break collision by construction (comparing node *types*, never rendered text) and the performance question at the same time (the standalone `O(m×n)` LCS-over-*lines* table is gone, replaced by two parses plus one tree walk, no worse asymptotically than what it replaced).

**Round 6 found the AST-pairing approach itself still had an unbounded cost, and a second, unrelated front-matter data-loss bug in round 5's own fix:**

- `pairChildrenByType` (the per-container node-type LCS) built a full `(m+1) × (n+1)` table unconditionally, on every call, even when — as in the dominant per-keystroke case (typing inside an existing paragraph changes no top-level node *types* at all) — the two children sequences are already identical. Fixed by trimming a common *suffix* first (walking both arrays backward from their ends, matching greedily while types agree) before falling back to the table for whatever's left. This is provably exactly equivalent to the untrimmed backtrace, not an approximation of it: the backtrace already starts at `(m, n)` and always takes a matching step first, unconditionally, whenever the trailing types agree, so replaying that same walk directly reproduces the identical matches. (A mirror-image *prefix* trim was considered and rejected — it is provably **not** equivalent, because a duplicate type run at the front (`[A, A, B]` vs `[A, B]`) can make the real backtrace match the *second* `A`, not the first, once the walk from the end reaches it. `source-line-map.test.ts` pins this exact counterexample against an independent untrimmed-LCS oracle.) Measured: `pairChildrenByType` on two identical 2000-element arrays went from ~28.4ms/call to ~0.083ms/call (~340×); full `generateUnifiedDiff` on a 2000-paragraph document with a one-paragraph edit (parsing/serialization dominated, per the honest-context note below) went from ~477ms to ~399ms.
- Round 5's fix for a *different* review finding (a `---`-fenced block containing only YAML comments, e.g. `# TODO: fill this in`, being incorrectly rejected as front matter) introduced its own silent-drop bug: `# Title\n## Subtitle` between fences is simultaneously valid as "nothing but comments" and as two ordinary ATX headings — an irreducible ambiguity, not a gap in enumeration. `normalizeWithFrontMatter`/`contentEqualsWithFrontMatter` treated the resulting `hasFrontMatter: true, data: null` shape as "nothing here to preserve," which was true when only genuinely-blank content reached that branch and became false once comment-only content could too — so a document like that would silently lose the `# Title`/`## Subtitle` text on normalize, and two documents differing only in that text would compare equal. Fixed by falling back to the parsed `raw` text (opaque bytes, not re-interpreted as YAML or Markdown) whenever `data` is null but `raw` isn't — see "Fix mechanism" above.

## Known limitation (flagging for the merge decision, not deferring)

This fixes the reported *line number*, not universal `git apply` fidelity. The hunk *body* is still rendered from the normalized documents (by design), so a hunk's line count still describes how many lines its own body shows, not how many raw source lines that region spans.

- When a blank-line collapse falls entirely before a hunk's start (the issue's own repro, and the common case with `contextLines: 0`), the mapped start is exact and `git apply --check` succeeds — verified empirically, see `unified-diff.test.ts`'s `hunk header line numbers reference source, not normalized text (cinder#1324)` block.
- When a collapse falls *inside* a hunk's displayed context window (default `contextLines: 3` for `generateUnifiedDiff`, `2` for `generateMarkdownSummary`), the source-mapped start and the normalized-space body can still disagree on span, and `git apply --check` can still reject the patch. Empirically, for `original = 'Alpha\n\n\n\nOriginal text\n'` with default `contextLines: 3`: the header comes out `@@ -1,3 +1,3 @@` (start correctly mapped to line 1), but `git apply --check` still rejects it, because the body's 3rd line ("Original text") is actually raw source line 5, not line 3.
- This is a structural consequence of diffing normalized text while reporting source-space coordinates — not a gap this fix left unclosed. Reporting the correct *number* for every line still narrows the original bug to exactly this case. Both changesets document this explicitly.
- Two more, deliberately-accepted limits from the AST-based mechanism itself: a node whose position span covers multiple lines (a table row, a multi-line paragraph) can't be split more finely than that span — interpolated within, exact at its boundaries; and if `parseOrThrow` ever fails on either document, alignment degrades to treating the whole document as unanchored rather than throwing.

Also flagging, not silently shipping: `@lostgradient/editor`'s ReviewEditor shows its front-matter panel (with a raw-YAML recovery field) only when `hasFrontMatter` is true. A document loaded with genuinely broken front-matter YAML (not user-typed through the panel itself — its raw-YAML editing path never commits invalid YAML into the document, so only externally-loaded/corrupted documents reach this) now renders that span as plain Markdown body content instead of surfacing the recovery field. No data is lost (`combineFrontMatterAndBody` still round-trips byte-for-byte, tested), but the recovery affordance moves from a dedicated field to inline editing. The same "no data lost, but the affordance moves" tradeoff now also applies to a comment-only front-matter block whose content turns out to actually be ATX headings (round 6, see above): it round-trips as part of the front-matter span rather than rendering as headings in the body.

## Performance (corrected from an earlier, wrong claim in this section)

An earlier version of this PR's changeset and this section claimed these exports were "user-triggered, not per-keystroke," and that the per-call cost therefore didn't matter the way a hot-path cost would. That claim was wrong, and round 4 is what disproved it: `ReviewEditor`'s hidden `formDiff`/`formSummary` inputs are `$derived` whenever the component has a `name`, so `buildSourceLineMap` runs on every content edit, not just a user-triggered "export" action.

Given that, the actual cost story is: two `parseOrThrow` calls plus one AST tree walk per `buildSourceLineMap` call, no worse asymptotically than the `O(m×n)` line-based LCS table this mechanism replaced (round 4) — and, as of round 6, the per-container node-type matcher inside that walk no longer builds its own `O(m×n)` table unconditionally either; it trims a common suffix first and only falls back to the table for whatever structural difference remains, which is empty in the dominant per-keystroke case (see "Revision history" above for the measured numbers). `buildSourceLineMapCached` (used by both `generateUnifiedDiff` and `generateMarkdownSummary`) additionally memoizes per exact `(source, normalized)` pair, so the `original` side — unchanged across most keystrokes — is a cache hit after the first call; only `current` is recomputed per keystroke.

Honest context on where the remaining cost actually goes: at realistic document sizes, parsing and serialization (`normalize()`, called on both `original` and `current` by `normalizeDocument`, independent of anything in this PR) and the pre-existing `computeLineChanges` line-diff in `unified-diff.ts` dominate the total per-call cost far more than the node-type matcher does post-round-6 — the matcher's round-6 fix narrows what was previously the single largest addition this PR made to that cost, it doesn't claim to be the dominant cost overall.

## Front matter: fixed by inventory, not by patching the two call sites a bot review named

Every consumer of `hasFrontMatter`/`parseFrontMatter` across `@lostgradient/markdown`, `@lostgradient/editor`, and `@lostgradient/chat` was audited (chat has none): `generateUnifiedDiff`'s `includeFrontMatter` prepend guard and `reviewStateToMarkdown` now check `fencePresent` (not `hasFrontMatter`) before prepending a new fence, since prepending only needs to know "does a fence already exist here," not "is it valid." The remaining consumers (`normalizeDocument`/`splitDocument`, `computeDiffWithFrontMatter`, `front-matter-fields.svelte`'s raw-YAML validation, and a hand-rolled `DiffViewer`-local implementation that never called the shared `parseFrontMatter` in the first place) only ever conditionally split or validate, never prepend, so `hasFrontMatter` stays the correct check for them.

Scoped to direct `hasFrontMatter`/`parseFrontMatter` consumers, since that's the actual risk this finding was about (a caller using the wrong boolean as a "safe to prepend" guard). `@lostgradient/cinder` (`packages/components`) also consumes `contentEqualsWithFrontMatter` (`change-tracker.svelte.ts`, for unsaved-changes detection), but never checks `hasFrontMatter` or `parseFrontMatter` directly, so it wasn't in scope for *this* finding -- it does benefit from the round-6 raw-preservation fix to `contentEqualsWithFrontMatter` itself, transparently.

## Round 7: the comment-only-YAML fix (round 6) had a UI-side consequence, found by two bots independently

`front-matter-fields.svelte`'s raw-YAML textarea (`ReviewEditor`'s front-matter panel) commits an
edit via `onchange(parsed.data)`. Once round 6 made `parseFrontMatter` correctly report
`hasFrontMatter: true` for comment-only content (`# TODO: fill this in`), that content's `parsed.data`
is still `null` -- the same `null` a genuinely-empty block also produces, and `data` alone can't tell
them apart. The parent (`replaceFrontMatterData`) received only that `null` and collapsed the block to
a bare `---\n---\n`, discarding whatever the user actually typed (editing `# TODO` to `# DONE` lost
the whole thing) with no error shown -- the textarea kept displaying the typed text until the next
reactive sync pulled the now-empty `raw` back down and blanked it.

Fixed by threading `parsed.raw` through as a second `onchange` argument (`FrontMatterFieldsProps.onchange`
is now `(data, raw?) => void`), and giving `replaceFrontMatterData` a `raw` parameter: when `data` is null
but `raw` is a non-blank string, it preserves `raw` verbatim instead of falling into
`stringifyFrontMatter`'s null-data collapse -- the same `data === null && raw !== null` pattern
`combineFrontMatterAndBody` already used elsewhere in the same file for the identical reason. Genuinely
empty/whitespace-only input still collapses correctly; only comment-only content changes behavior.

## Release scope: three packages, not two

`@lostgradient/markdown` bumps minor (the `fencePresent` addition and the `hasFrontMatter` breaking-narrowing are both handled as a minor, pre-1.0). `@lostgradient/editor`'s peer range on `@lostgradient/markdown` (currently `^0.2.0`) doesn't cover `0.3.0`, and neither does `@lostgradient/chat`'s — both need widening, or the release breaks their own consumers. Rather than hand-editing `peerDependencies` in this PR (verified empirically to break `packages/editor/scripts/package-boundary.test.ts`, since that guard checks the peer range against the dependency's *current*, not future, published version), this uses the repo's own existing mechanism: `.changeset/reconcile-internal-peers.md`, generated by `packages/components/scripts/prepare-internal-peer-changesets.ts`, which the release pipeline's `reconcile-internal-peers.ts` step widens for real, post-version-bump, once `@lostgradient/markdown`'s `package.json` version has actually changed. `bunx changeset status` confirms all three packages (`@lostgradient/markdown`, `@lostgradient/editor`, `@lostgradient/chat`) are staged to release together.

## Revert verification (every fix has a test proven to fail without it)

- Reverting `packages/markdown/src/pipeline/frontmatter.ts` (and, for the round-6 addition, `ast.ts`) individually: each targeted test goes red for exactly the case it pins, with all other markdown tests staying green. Restored and re-verified green each time, confirmed byte-identical to a backup by hash.
- Reverting `packages/editor/src/lib/export/{unified-diff,markdown-summary,source-line-map}.ts` together (round 1 shape) and again at each subsequent round's shape: the corresponding new tests go red each time (the Setext/thematic-break collision, the comment-only front-matter drop, etc.); the `normalizeInputs: false` identity-map tests correctly stay green either way, since they pin behavior that was never broken. Restored and re-verified green each time, confirmed byte-identical to backups by hash.
- The round-6 suffix-trim optimization is a pure performance change with identical output by construction — there is no test that fails without it (a timing assertion would be the threshold-padding this repo's own rules ban). Its test coverage is an equivalence check against an independent untrimmed-LCS oracle across a shape-varied corpus (including the prefix-trim counterexample above), which exists to catch a *future* regression in the trim's soundness, not to prove the optimization currently exists. The optimization's actual effect is reported as measured numbers in "Revision history" instead.
- Reverting `front-matter-fields.svelte`, `review-editor-front-matter.ts`, and `review-editor-impl.svelte` (round 7) to their pre-fix state: exactly 4 tests go red (the comment-only data-loss repro, plus two existing `onchange` assertions updated to expect the new second argument), all 18 other tests across both touched test files stay green. Restored and re-verified green, confirmed byte-identical to a backup by hash.

## Test plan

- [x] `bunx turbo run test typecheck lint --filter=@lostgradient/markdown --filter=@lostgradient/editor --filter=@lostgradient/chat --force` — all green (`@lostgradient/markdown`: 1068 pass, 26 pre-existing skip; `@lostgradient/editor`: 829 pass, 1 pre-existing skip; `@lostgradient/chat`: 779 pass); 0 typecheck errors, 0 lint warnings/errors across all three
- [x] `bun run --filter=@lostgradient/editor components:check` — OK
- [x] `bun run --filter=@lostgradient/cinder check:changeset-prerelease-bumps` — OK
- [x] `bunx prettier --check` on every touched file — OK
- [x] `bunx changeset status` — recognizes all three changesets (`@lostgradient/markdown`, `@lostgradient/editor`, `@lostgradient/chat`, all minor)
- [x] Every fix above independently revert-and-restore verified, each restore hash-confirmed against a backup

Fixes #1324
Fixes #1325


stevekinney:
All five bot review findings addressed for real (fixed, tested, break-and-restore verified), not dismissed:

1. **copilot** — `splitLines()` dropping a single-newline document to zero lines: fixed. Removed the erroneous "stripped-to-empty means zero lines" special case; `'\n'` now correctly splits to `['']` (one blank line), matching `splitIntoLines` in `unified-diff.ts`.

2. **copilot** — hunk header mixes coordinate spaces (source-space start, normalized-space count) with no explanation: documented, both inline at the push site and in `generateUnifiedDiff`'s own docblock. This is intentional (the body is still rendered from normalized text, so the count has to describe the body it's attached to), not a bug to "fix" by making the count disagree with its own body — see the new "Known limitation" docblock section.

3. **codex, P1** — rewritten lines (e.g. a collapsed Setext heading) fell back to the *preceding* matched line instead of their own position: fixed. `fillUnmatchedRuns` now interpolates forward from the preceding match instead of freezing on it, clamped to the next real match so the map stays monotonically non-decreasing even across a long unmatched run between two closely-spaced matches. Verified against the exact repro from the review comment (`Intro\n\nOld title\n===`): now reports source line 3, not line 2.

4. **codex, P2** — `mapNormalizedLineNumber` clamped past-EOF lookups onto the last real line, losing the "insert after the last line" position for a pure trailing addition: fixed. Past-the-end lookups now extrapolate past the last mapped source line. Verified against the exact repro (`normalizeInputs: false`, `original: 'Alpha'`, `current: 'Alpha\nBeta'`, `contextLines: 0`): now reports `### Lines 2-2`, not `### Lines 1-1`.

5. **codex, P2** — the exported `hasFrontMatter()` helper contradicted `parseFrontMatter()` for the exact false-positive inputs the #1325 fix targets, despite its own doc comment promising consistency: fixed. Now delegates to `parseFrontMatter()` instead of re-deriving a shallower check.

Every fix has a dedicated test proven to fail without it: reverted to the pre-review-fix commit (`dfce8d713`), confirmed all 8 new/updated tests go red, restored, confirmed green, verified byte-identical to a backup by `diff`. Full `turbo run test typecheck lint` for both `@lostgradient/markdown` and `@lostgradient/editor` green on the new commit (`68a8ff145`).

stevekinney:
Follow-up finding (chatgpt-codex-connector, P2) fixed: verified empirically first (the exact repro reported `### Lines 2-2` instead of the correct `### Lines 4-4`), real bug, real fix.

`buildSourceLineMap`/`identitySourceLineMap` now return a `SourceLineMap` (`{ lines, sourceLineCount }`) instead of a bare `number[]`, and `mapNormalizedLineNumber` extrapolates past-the-end lookups from `sourceLineCount` — the source document's own true line count — rather than from the normalized document's line count or the map's last entry. Those are genuinely different numbers whenever normalization strips trailing content *entirely* (e.g. blank lines at EOF, which have nothing after them to anchor a representative line the way a mid-document blank run does) rather than collapsing it to one representative line.

Also confirmed, and fixed a misleading comment over: `generateUnifiedDiff` itself never actually reaches this extrapolation branch for hunk headers — `buildHunk`'s own "start=0 when count=0" git convention bypasses the line-map lookup entirely for a pure-addition hunk with no original-side content, so its existing "insert after EOF" test was passing via the ordinary in-bounds path even before this fix, not exercising the branch its comment claimed to. Fixed the comment to say so rather than leave it implying coverage it didn't have. The extrapolation mechanism is properly exercised by `generateMarkdownSummary`'s own line-counting (which reaches it directly) and by dedicated unit tests in `source-line-map.test.ts`.

Verified failing against the pre-fix commit (`68a8ff145`) — 14 assertions red, including the key `Lines 2-2` vs `Lines 4-4` behavioral check — restored, confirmed green, byte-verified by diff. Full `turbo run test typecheck lint` for `@lostgradient/editor` green on the new commit (`487a2ac40`).

stevekinney:
Two more findings from this review round, both real, both fixed:

1. **codex, P2** (`source-line-map.ts`) — Verified the exact repro empirically first (`Intro\n\n* one\n\n* old` -> normalized `Intro\n\n- one\n- old` produced map `[1,2,3,4]`, mapping the second item to the deleted blank separator's line instead of its own source line 5). Fixed by canonicalizing unordered list markers (`-`/`*`/`+`) before the LCS equality check in `alignNormalizedToSource`, not by further patching the interpolation fallback — a marker-only change is now recognized as the *same* line surviving (consuming its own source line correctly), which leaves the genuinely-deleted blank separator correctly unrepresented in the map, rather than trying to make the interpolation heuristic guess correctly around a deletion it has no way to detect. Verified: map is now `[1,2,3,5]`; editing the second item with `contextLines: 0` now reports line 5 in both `generateUnifiedDiff` and `generateMarkdownSummary`.

2. **codex, P2** (really `front-matter-fields.svelte:86-94`, not `frontmatter.ts` itself, per the coordinator's framing) — Confirmed `validateFrontMatter()`'s actual implementation first: it only checks that content parses as YAML at all, no object-shape check. Reproduced the silent discard exactly as described (typing `- one` into the raw-YAML field of an empty front-matter block: `onchange(null)` fires, the parent round-trips back to the original empty block via `preserveEmptyFrontMatter`, no error shown, textarea keeps displaying the discarded input). This is a distinct, more consequential bug class than the line-mapping refinements — genuine silent data loss in a UI commit path, exposed by #1325 making `parseFrontMatter` stricter without this consumer keeping up. Fixed by having `handleRawInput` confirm `parseFrontMatter(...).hasFrontMatter` before committing (not just `validateFrontMatter`'s syntax check), surfacing a validation error instead of discarding silently. New `front-matter-fields.test.ts` mounts the component with `@testing-library/svelte` and drives the exact repro via simulated user input.

Both verified failing against the pre-fix commit (`487a2ac40`) — 5 assertions red across `source-line-map.test.ts`, `unified-diff.test.ts`, `markdown-summary.test.ts`, and the new `front-matter-fields.test.ts` — restored, confirmed green, byte-verified by diff. Full `turbo run test typecheck lint` for `@lostgradient/editor` green on the new commit (`c50107775`).

stevekinney:
Resolved the ordered-marker finding (`packages/editor/src/lib/export/source-line-map.ts`) — and stepped back from patching it as a fourth instance of the same bug shape.

**What changed (commit `de41787a9`):** this is round 4 of the identical failure — Setext folding, unordered markers, now ordered markers, each fixed by hand-encoding one more rewrite kind into `linesMatchForAlignment`/`canonicalizeListMarker`. That's re-deriving a copy of what `@lostgradient/markdown`'s `normalize()` already does — the same "two normalizers drift apart" class `normalizeDocument()`/`splitDocument()` (cinder#1307, cinder#1318) fixed once already, for front-matter handling.

Replaced the hand-rolled canonicalizer with `canonicalizeLine`, which calls the real `normalize()` per line (memoized per `buildSourceLineMap` call, cache scoped to that one call) instead of re-deriving rewrite rules. This covers every member of `serializerOptions` — both list marker kinds, emphasis, strong, thematic-break rule characters, list-item indent spacing — by construction, plus whatever `remark-stringify`/`remark-gfm` defaults it doesn't override. Two documented guards keep a per-line call safe: blank/whitespace-only lines short-circuit to `''`, and a canonicalization result containing a newline (a multi-line-context transform like Setext folding, misread indented-code re-fencing) is discarded, falling back to the original line rather than risking a false match.

**Regression tests added**, each verified to *actually fail* against the previous marker-only implementation (not just asserted to pass — first draft of this block used constructions that passed either way, caught before committing):
- ordered markers (`1)` → `1.`) — the round-4 repro itself
- emphasis, strong, and list-item indent spacing, each embedded in tight-list content so the zero-residue separator-deletion mechanism applies the same way it did for unordered markers
- thematic break (`***` → `---`), using a different unambiguous construction since it can't be list-item content: `normalize()` inserts blank padding around a bare rule with no blanks of its own to create alignment slack

All prior rounds' repros (Setext heading fold, unordered bullets `*`/`+`, both EOF/trailing-content cases) remain committed and passing, unchanged.

Full break-and-restore output: reverting to the marker-only implementation turns exactly these 5 new tests red (ordered marker + the 4 inventory tests) while all 15 prior tests — including the round-3 unordered-marker test — stay green, confirming the old fix wasn't wrong, just incomplete. Restored and hash-verified identical (`sha256sum` match) before re-running green.

`bunx turbo run test typecheck lint --filter=@lostgradient/markdown --filter=@lostgradient/editor` and `components:check` all pass. Changeset updated to describe the architectural change rather than the marker-by-marker history.

stevekinney:
Follow-up to my previous comment (commit \`de41787a9\`), pushed as \`bf833991c\`:

**Correction to my own inventory claim.** I wrote "this covers every member of \`serializerOptions\` — both list marker kinds, emphasis, strong…" — that's wrong about the ordered marker specifically. Ordered-list style (\`1)\` vs \`1.\`) comes from \`remark-stringify\`'s own \`bulletOrdered\` default, not from a \`serializerOptions\` entry — \`serializerOptions\` only sets \`bullet\` (the unordered marker). That's exactly why the ordered-marker case was its own review-round finding rather than falling out of the unordered-marker fix "for free": nothing in \`serializerOptions\` names it. \`canonicalizeLine\` still covers it correctly (it calls the real \`normalize()\`, which reaches every remark-stringify/remark-gfm default whether or not this file names it), so the *fix* was never wrong — only my inventory description of *why* was imprecise. Corrected the attribution in the module docblock and the changeset.

Same pass also named two actual, deliberate carve-outs I'd only implied before: \`fence\`/\`fences\` and \`setext\` are real \`serializerOptions\` members, but their rewrites (a Setext underline folding into the heading above it, \`~~~\` becoming a fenced *pair*) need multi-line context a single isolated line can't reproduce — they hit \`canonicalizeLine\`'s multi-line guard and fall back to the original line, leaving those two cases exactly as approximate as they were before this fix. Not a gap this fix silently leaves; stated plainly in both the docblock and the changeset now, and it's the same guard the pre-existing Setext test already exercises (not separately tested for fence-style, since a construction that passes whether or not the guard exists isn't a real regression pin — verified that empirically before deciding not to add one).

Also added the round-4 (ordered-marker) repro at the consumer level — \`unified-diff.test.ts\` and \`markdown-summary.test.ts\`, mirroring the parallel pins round 3's unordered-marker finding already got in both files, alongside the \`source-line-map.test.ts\` unit test. Both verified red against the round-3 (marker-only) implementation, then restored and hash-verified identical.

\`bunx turbo run test typecheck lint --filter=@lostgradient/markdown --filter=@lostgradient/editor --force\` (788 editor tests, 1059 markdown tests, typecheck clean) and \`components:check\` all pass on \`bf833991c\`.

stevekinney:
Resolved all five round-5 threads (commit \`772c38ab3\`). Two of them (the Setext/thematic-break string collision and the reactive-per-keystroke performance finding) required an architecture change, not another patch — summarizing what changed and why, plus the two front-matter findings and the peer/changeset situation.

**Architecture: AST-based alignment, replacing string canonicalization entirely.**

The collision finding was conclusive: \`Title\n---\` (a Setext heading) and \`***\` (a thematic break) are different mdast node types that can both canonicalize to the literal string \`---\`. Verified directly — \`buildSourceLineMap('Title\n---\n***\nAfter\n', normalize(...))\` mapped the emitted rule (source line 3) to the deleted Setext underline (source line 2) under the previous per-line-\`normalize()\` implementation. No amount of per-line text canonicalization can fix this: once two structurally different nodes serialize to the same string, the string alone can't say which one it came from.

\`source-line-map.ts\` now parses both \`source\` and \`normalized\` (\`parseOrThrow\`, \`@lostgradient/markdown/pipeline\`) and walks both mdast trees in lockstep, pairing corresponding nodes by structural \*type\* (an LCS over node types, replacing the LCS over line strings) rather than comparing rendered text. Each paired leaf node contributes an exact anchor from its own source/normalized position spans, linearly interpolated when the spans differ in length (a Setext heading's two-line source span folding into its one-line ATX form, for instance — verified this still resolves to the *start* of the source span, matching the pre-existing test). Gaps between anchors still use the same forward-interpolation-clamped-to-next-anchor fallback as before, just anchored by node boundaries now instead of string matches. This makes the collision impossible by construction: a \`heading\` node and a \`thematicBreak\` node are never "the same node" regardless of what text they happen to serialize to.

This also resolves the performance finding directly rather than disputing it — confirmed by reading \`review-editor-impl.svelte\`: \`formDiff\`/\`formSummary\` are \`$derived\` and rendered inside \`{#if name}\`, so \`exportUnifiedDiff\`/\`exportMarkdownSummary\` really do re-run on every content edit whenever the component has a \`name\`, not just on a user-triggered export (a claim an earlier round's changeset made incorrectly — corrected). The rewrite removes the \`O(m×n)\` LCS-over-lines table and the per-line \`normalize()\` calls entirely (two parses + one tree walk instead), and \`buildSourceLineMapCached\` adds an LRU cache over \`(source, normalized)\` pairs — mirroring \`@lostgradient/markdown\`'s own \`normalizeWithCache\` — wired into both \`generateUnifiedDiff\` and \`generateMarkdownSummary\`, so the \`original\`-side map (unchanged across most keystrokes, and shared between the two sibling exports) is a cache hit after the first call.

**Verification of the architecture's soundness precondition**, not just assertion: tree-shape-stability tests confirm \`source\`'s and \`normalized\`'s top-level node-type sequences match 1:1 across the full repro corpus (identity, blank-collapse, Setext fold, both marker kinds, the collision case, emphasis/strong/indent-in-list, bare thematic break, fence style) — the precondition the lockstep pairing relies on.

**All six repros are committed regression tests**, each verified to fail against the prior (round-4, per-line-\`normalize()\`) implementation before being restored and re-verified green: Setext fold, unordered/ordered marker deletion (both already covered pre-round-5), emphasis/strong/list-item-indent inside tight lists, a bare thematic break with synthesized padding, and the new Setext/thematic-break collision — pinned at the unit level (\`source-line-map.test.ts\`) and, since the codex finding named both, at the \`generateUnifiedDiff\`/\`generateMarkdownSummary\` consumer level too.

**Front matter, fixed by inventory:**

- Comment-only fences (\`---\n# note\n---\`): \`load()\` returns \`null\` for this, identical to genuinely blank content, so it fell into the "reject" branch alongside a real sequence/scalar. \`isCommentOnlyYaml\` (every line trimmed is blank or starts with \`#\`) routes it to the same "empty, still front matter" path the whitespace-only case already had.
- Wrapped fences: \`FrontMatterParseResult\` gains a third state, \`fencePresent\` — true whenever a \`---\`...\`---\` span exists at all, regardless of validity, distinct from \`hasFrontMatter\` (valid only). Grepped every consumer of \`hasFrontMatter\`/\`parseFrontMatter\` across \`packages/markdown\` and \`packages/editor\` (chat has none): the two "is it safe to prepend a new fence" call sites (\`generateUnifiedDiff\`'s \`includeFrontMatter\`, and \`reviewStateToMarkdown\`, the latter not named in the original finding — found by the grep) now check \`fencePresent\`; every other consumer only ever conditionally splits or validates, never prepends, so \`hasFrontMatter\` stays correct there. Consumer-level double-wrap regression tests added for both call sites.

**Peer ranges / changeset scope:** confirmed the P1 finding (\`^0.2.0\` excludes the \`0.3.0\` this PR ships) — but rather than hand-editing \`peerDependencies\` in this PR, used the repo's own established mechanism: \`bun run packages/components/scripts/prepare-internal-peer-changesets.ts\`, which detected the pending Markdown minor and wrote \`.changeset/reconcile-internal-peers.md\` bumping both \`@lostgradient/chat\` and \`@lostgradient/editor\` to minor. Verified hand-editing the range now would actually *break* CI: \`package-boundary.test.ts\`'s own guard checks the peer range against \`packages/markdown/package.json\`'s *current* (not-yet-bumped) version, which is still 0.2.0 until the release PR runs \`reconcile-internal-peers.ts\` post-version-bump — tried it and watched two tests fail, then reverted. \`bunx changeset status\` now shows a three-package minor release (chat, editor, markdown).

\`bunx turbo run test typecheck lint --filter=@lostgradient/markdown --filter=@lostgradient/editor --filter=@lostgradient/chat --force\` (811 editor, 1062 markdown, 779 chat tests), \`components:check\`, and \`check:changeset-prerelease-bumps\` all pass on \`772c38ab3\`.

stevekinney:
Round-6 finding (codex, P2) fixed: **the quadratic AST child matcher**.

Verified the concern first, not just the claim: measured `pairChildrenByType` directly on two identical 2000-element type arrays (simulating a large document's top-level children with no structural edit -- exactly what typing inside an existing paragraph produces, since it changes no top-level node *type*). Before: ~28.4ms/call. After: ~0.083ms/call. ~340x.

**Fix**: trim a common *suffix* first -- walk both arrays backward from their ends, matching greedily while `nodeKey`s agree -- before falling back to the `O(m×n)` table for whatever's left. This is provably *exactly* equivalent to the untrimmed backtrace, not an approximation: the backtrace already starts at `(m, n)` and unconditionally takes a matching step first whenever the trailing types agree, so replaying that same walk directly reproduces the identical matches for that run.

I also considered the mirror-image *prefix* trim (walk forward from index 0) and rejected it -- it is **not** equivalent. Worked counterexample: `source = [A, A, B]`, `normalized = [A, B]`. The real backtrace matches normalized's lone `A` against source's *second* `A`, not the first, because by the time the walk (starting from the end) reaches index 0 it's already consumed the first `A` via a different path. A naive prefix trim would match the first `A` instead -- a genuinely different, wrong answer. Committed this exact case as a regression test (`packages/editor/src/lib/export/source-line-map.test.ts`, `pairChildrenByType suffix-trim optimization`) against an independent, deliberately-not-imported untrimmed-LCS oracle, across 12 shape-varied and shape-mismatched cases.

Being upfront about test coverage here: this is a pure optimization with identical output by construction, so there's no test that fails without it -- a timing assertion would be the threshold-padding this repo's rules explicitly ban. The committed tests are an equivalence check (guards against a *future* regression in the trim's soundness), not proof the optimization exists. That proof is the benchmark numbers above, run manually, not committed.

Honest context on end-to-end impact: full `generateUnifiedDiff` on a 2000-paragraph document with a one-paragraph edit (original-side cache warm) went from ~477ms to ~399ms -- a real ~16% improvement, but far short of 340x, because parsing/serialization and the pre-existing `computeLineChanges` line-diff dominate the total at this scale, not the matcher. The asymptotic fix matters most as documents get larger relative to those fixed costs; PR body's "Performance" section now says this explicitly instead of overclaiming.

Break-and-restore: reverted `pairChildrenByType` to the round-5 (`772c38ab3`) unconditional-table shape, confirmed all 48 tests in `source-line-map.test.ts` still pass (expected -- the equivalence tests check output identity, not which code path produced it), then confirmed via direct benchmarking that the untrimmed version is genuinely ~340x slower on the matcher before restoring. Restore hash-verified against a backup.

stevekinney:
Round-6 finding (Cursor Bugbot, Medium -> treating as real, not a nitpick) fixed: **comment-only fences drop body text**.

Confirmed the exact repro first: `---\n# Title\n## Subtitle\n---\nBody content\n` -- `isCommentOnlyYaml` (added in round 5, to fix a *different* review finding: comment-only YAML front matter was being incorrectly rejected) classifies `# Title\n## Subtitle` as "nothing but comments," so `parseFrontMatter` reports `hasFrontMatter: true, data: null` -- the same shape a genuinely-blank block reports. `normalizeWithFrontMatter`'s `!data` branch returned only `normalize(body)`, silently discarding the `# Title\n## Subtitle` span entirely. `contentEqualsWithFrontMatter` had the mirror bug: two documents differing only in that text compared equal, since `data` is `null` on both sides regardless of what the comment text says.

This is real, and it's the second time this exact failure mode (front-matter-shaped span silently dropped or made invisible to comparison) has shown up in this PR -- round 5's `fencePresent` fix closed one instance of it, and round 5's own comment-only fix reopened a new one.

**Root cause, and why I didn't just revert the comment-only classification**: `# Title\n## Subtitle` between fences is *irreducibly* ambiguous -- it's simultaneously valid as "nothing but YAML comments" and as two ordinary ATX headings sandwiched between thematic breaks. There's no object-shape test that resolves this the way there is for a list vs. a YAML sequence, because neither reading produces YAML data. Reverting the comment-only fix would regress the round-5 finding it addressed; the actual defect isn't the classification, it's that `normalizeWithFrontMatter`/`contentEqualsWithFrontMatter` treated "no data" as "nothing to preserve or compare," which was true when only genuinely-blank content could reach that branch and became false once comment-only content could too.

**Fix**: both functions now fall back to the parsed `raw` text (the literal bytes between the fences) whenever `hasFrontMatter` is true, `data` is null, and `raw` is non-null -- `normalizeWithFrontMatter` re-wraps it verbatim in the output instead of discarding it; `contentEqualsWithFrontMatter` compares it byte-for-byte before falling through to the body comparison. `raw` is treated as opaque text either way (never re-interpreted as YAML or Markdown), so misclassifying ATX headings as a comment-only block now costs a display affordance at most (they round-trip as part of the front-matter span rather than rendering as headings), never the underlying bytes -- the same "no data lost, affordance moves" tradeoff this PR already established for genuinely-invalid front matter.

One deliberately-accepted, documented edge case from this fix: `---\n{}\n---` (`raw: '{}'`, an empty YAML mapping) no longer compares equal to `---\n---` (`raw: null`) via `contentEqualsWithFrontMatter`, even though both have `data: null` -- a spurious "changed" rather than a missed one, the safe direction to be wrong on. Pinned as its own test so a later round doesn't rediscover it as a mystery.

Also updated `isCommentOnlyYaml`'s docblock, which previously (and now, per this finding, incorrectly) asserted "a document doesn't ordinarily have `# ...` as its very first line followed immediately by a second `---`" -- that's exactly the case this finding falsified.

Break-and-restore: reverted `packages/markdown/src/pipeline/ast.ts` to the round-5 (`772c38ab3`) shape, confirmed exactly the 3 new tests go red (the span-drop, the false-equal, and the `{}`-vs-blank edge case) with all 55 prior tests staying green, restored, hash-verified identical, re-confirmed all 58 green.

stevekinney:
Round-7 finding (chatgpt-codex-connector and copilot-pull-request-reviewer, independently, same bug) fixed: **`FrontMatterFields`' comment-only commit silently collapsed to an empty fence**.

Confirmed the mechanism first, exactly as described: `handleRawInput` calls `onchange(parsed.data)`. For comment-only YAML (`# TODO: fill this in`), the round-5/6 fix correctly makes `parsed.hasFrontMatter` `true`, so the earlier "reject invalid input" gate doesn't fire -- but `parsed.data` is still `null` for that case, the same `null` a genuinely-empty block produces. The parent (`replaceFrontMatterData`) received only that `null` and collapsed the block to `---\n---\n`, discarding whatever the user typed (`# TODO` -> `# DONE` loses the whole thing) with no error shown.

**Fix**: `data` alone can never distinguish "recognized front matter with no data" from "genuinely empty" -- both produce `null`. Threaded `parsed.raw` through as a second `onchange` argument (`FrontMatterFieldsProps.onchange` is now `(data, raw?) => void`), forwarded through `handleFrontMatterChange`, and `replaceFrontMatterData` gained a `raw` parameter: when `data` is null but `raw` is a non-blank string, it now preserves `raw` verbatim instead of falling into `stringifyFrontMatter`'s null-data collapse -- the exact same `data === null && raw !== null` pattern `combineFrontMatterAndBody` already used elsewhere in this same file for the identical reason (I didn't invent a new pattern, I applied the existing one to the path that was missing it). Genuinely empty/whitespace-only input still collapses correctly.

Break-and-restore: reverted `front-matter-fields.svelte`, `review-editor-front-matter.ts`, and `review-editor-impl.svelte` to the pre-fix (`c754f8822`) state, confirmed exactly 4 tests go red -- the data-loss repro itself, plus two existing `onchange` assertions updated to expect the new second argument -- with all 18 other tests across both touched test files staying green. Restored and hash-verified byte-identical to a backup, re-confirmed all 22 green.

Full `@lostgradient/editor` suite (829 pass, 1 pre-existing skip, 0 fail), typecheck, lint, and `components:check` all green on the fix.

PR #1335, opened 2026-08-15
fix(editor): replace validate-consumer's import-closure regex with a real lexer

### LINKED ISSUE #1334: editor: validate-consumer's import-closure regex false-positives on prose in preserved doc comments
## Summary

`packages/editor/scripts/validate-consumer.ts`'s `assertImportClosure` scans every emitted `dist/**/*.{js,svelte,css}` file for import specifiers using two regexes:

```ts
const patterns = [
  /(?:\bfrom\s*|\bimport\s*\()\s*['"]([^'"]+)['"]/gu,
  /@import\s+['"]([^'"]+)['"]/gu,
];
```

The first pattern matches the literal word `from` followed by optional whitespace and a quote character — it does not require an actual `import ... from '...'` statement. Preserved doc comments in the emitted `.js`/`.svelte` files are plain text, and when a comment happens to contain the word "from" followed (after whitespace) by a straight quote character used for prose (a scare-quote, a contraction, an em-dash-adjacent quote), the regex captures everything up to the next quote character as if it were an import specifier.

## Repro

This first surfaced in the release workflow for PR #1330/#1332 (`chore: version packages`), which failed at the `Validate Editor package artifact` step:

```
error: [validate-consumer] packed production imports are not declared peers:
  dist/components/review-editor/review-editor-front-matter.js: genuinely empty, safe to
        // collapse
  dist/components/review-editor/front-matter-fields.svelte: genuinely
     * empty, safe to collapse
  dist/export/source-line-map.js:  for every node the
 * parser understood, with no string comparison anywhere.
  dist/export/source-line-map.js:  behavior for a Setext heading
```

None of the captured "specifiers" are real import paths — they're fragments of prose from doc comments in `source-line-map.ts` and `front-matter-fields.svelte`, both added/expanded by #1330's AST-native line-provenance rewrite. `bun run --filter=@lostgradient/editor validate:consumer` reproduces this at current `origin/main`.

## Why this was never caught before

`validate:consumer` only runs in the release workflow, not in PR CI — so five straight green CI passes on #1330 never exercised it. This is the third instance in the same campaign of a string-heuristic standing in for something that needed a real parser (the LCS line-map reconstruction, the hardcoded peer-range literal in #1333, now this).

## Requested fix

Replace the regex-based scan with a real import lexer for `.js` files — `Bun.Transpiler.scanImports()`, since the validator already runs under Bun — rather than tightening the regex further. `.svelte` script blocks and CSS `@import` may need their own handling; a strict specifier-shape guard (reject captures containing whitespace) is an acceptable fallback there if a full lexer isn't practical for those file types.

Do not fix this by rewording the doc comments that happen to trigger it — that treats the symptom, not the validator's actual bug, and the next prose-heavy comment will trigger it again.

stevekinney:
## Summary

Fixes #1334. `packages/editor/scripts/validate-consumer.ts`'s `assertImportClosure` scanned packed `dist/**/*.{js,svelte,css}` for bare import specifiers using `/(?:\bfrom\s*|\bimport\s*\()\s*['"]([^'"]+)['"]/gu` — which matches the literal word "from" followed by a quote character anywhere in the file, with no requirement that it's part of an actual `import ... from '...'` statement. #1330's AST-provenance rewrite added prose-heavy doc comments to `source-line-map.ts` and `front-matter-fields.svelte` that happen to contain "from" followed by a quote (a scare-quote, a contraction), and the regex captured that prose as if it were an import specifier — blocking the real release workflow's npm publish, with zero relationship to any actual undeclared import.

## Reproduced first, read in full

```
bun run --filter=@lostgradient/editor validate:consumer
```

fails at current `origin/main` with exactly 4 flagged "specifiers" — read every one before concluding they're false positives, not just the tail-truncated release log:

```
dist/export/source-line-map.js:  for every node the
 * parser understood, with no string comparison anywhere.
 ...
dist/export/source-line-map.js:  behavior for a Setext heading
dist/components/review-editor/review-editor-front-matter.js: genuinely empty, safe to
      // collapse
dist/components/review-editor/front-matter-fields.svelte: genuinely
   * empty, safe to collapse
```

None are real import paths — all four are prose fragments from doc comments in the two files #1330 touched most.

## Fix

Not a tighter regex, and not rewording the comments that happen to trigger it — the fix belongs in the validator, per the issue. Uses `Bun.Transpiler.scanImports()` for `.js` files: a real lexer, structurally blind to comment text since it tokenizes the file the way it would to actually run it. This mirrors an **existing, unrelated precedent already in this repo** — `packages/markdown/scripts/validate-consumer.ts`'s own `assertImportClosure` already fixed this exact bug class for its `.js`-only case; this PR applies the same fix to Editor's larger surface, which markdown's doesn't have:

- **`.js`**: `Bun.Transpiler.scanImports()` directly.
- **`.svelte`**: no Svelte-markup-aware lexer is available, so each `<script>` block's content (module and instance) is extracted first, then the same real lexer runs on just that — markup and `<style>` content are never scanned for imports at all.
- **`.css`**: no import-statement lexer available either, so `@import` detection stays regex-based, guarded by a strict specifier-shape check that rejects any capture containing whitespace — which rules out comment prose specifically, since no real import specifier can ever contain a space.

## Verification (both directions, not just that the false positives go away)

A validator that can't fail is worse than no validator, so `packages/editor/scripts/validate-consumer.test.ts` (new) exercises the real, now-exported `assertImportClosure` against real fixture files on disk — not a reimplementation of its logic:

- Comment prose containing "from" + a quote (the exact repro shape, for `.js` and `.svelte`) never triggers a violation.
- A genuine undeclared bare import — in a `.js` file, inside a `.svelte` `<script>` block, and via a real CSS `@import` — is still caught and rejected in all three cases.
- Markup and `<style>` content outside any `<script>` tag is never scanned at all.
- `extractSvelteScriptBlocks` and the specifier-shape guard are also covered directly with focused unit tests.

**Break-and-restore**: swapped `assertImportClosure`'s body back to the pre-fix regex logic (keeping the new exports so the test file still loads) — exactly the 4 "prose is not an import" tests go red, every "a real import is still caught" test and every pure-helper test stays green regardless (they never depended on the bug). Restored and hash-verified byte-identical to a backup, re-confirmed all 22 tests green.

Full `bun run --filter=@lostgradient/editor validate:consumer` now completes successfully end-to-end (build, import closure, client build, plugin SSR, plain-Node SSR, svelte-check) — this is the actual release-workflow step that was failing.

## Changeset

None. Scripts-only, no package runtime behavior changes; `bun packages/components/scripts/check-changeset-prerelease-bumps.ts` passes with zero changesets present (verified locally, same check confirmed for #1333).

## Scope note

Per the issue: this does **not** add `validate:consumer` to PR CI (it currently only runs in the release workflow, which is why 5 green PR-CI passes on #1330 never caught this). That's a pipeline change, flagged in #1334 for a separate decision, not made here.

## Test plan

- [x] Reproduced #1334's failure at current `origin/main`, read the full violation list (4 entries, all prose)
- [x] Fix confirmed: `bun run --filter=@lostgradient/editor validate:consumer` completes successfully end-to-end
- [x] New `validate-consumer.test.ts`: 13 tests, both directions (false positives eliminated, real undeclared imports still caught)
- [x] Break-and-restore: reverting to the old regex logic makes exactly the 4 relevant tests fail; restore hash-verified
- [x] `bunx turbo run test typecheck lint --filter=@lostgradient/editor --force` — all green (842 pass, 1 pre-existing skip), 0 typecheck/lint errors
- [x] `bunx prettier --check` — OK
- [x] `bun packages/components/scripts/check-changeset-prerelease-bumps.ts` — OK, no changeset needed

### DIFF
--- a/packages/editor/scripts/validate-consumer.test.ts
+++ b/packages/editor/scripts/validate-consumer.test.ts
@@ -1,12 +1,14 @@
 import { afterEach, describe, expect, test } from 'bun:test';
 import { mkdir, mkdtemp, rm } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
-import { join } from 'node:path';
+import { dirname, join } from 'node:path';
 
 import type { PackageManifest } from './pack-for-publish.ts';
 import {
   assertImportClosure,
   extractSvelteScriptBlocks,
+  extractSvelteStyleBlocks,
+  isNodeBuiltinSpecifier,
   isPlausibleImportSpecifier,
 } from './validate-consumer.ts';
 
@@ -43,7 +45,14 @@ describe('assertImportClosure (cinder#1334)', () => {
     tempRoots.push(root);
     for (const [relativePath, content] of Object.entries(files)) {
       const fullPath = join(root, relativePath);
-      await mkdir(join(fullPath, '..'), { recursive: true });
+      // `dirname()`, not `join(fullPath, '..')`: both happen to resolve to
+      // the same parent directory here (`path.join`'s lexical normalization
+      // cancels a trailing `/<segment>/..` regardless of whether `<segment>`
+      // looks like a file or a directory name -- verified empirically, this
+      // was never actually broken), but `dirname()` says what it means
+      // instead of relying on a normalization side effect a reviewer has to
+      // re-derive to trust (review finding, cinder#1335 round 2).
+      await mkdir(dirname(fullPath), { recursive: true });
       await Bun.write(fullPath, content);
     }
     return root;
@@ -133,7 +142,7 @@ export const value = 1;
     );
   });
 
-  test('.svelte files: markup and <style> content outside <script> is never scanned at all', async () => {
+  test('.svelte files: markup outside <script> and <style> is never scanned at all', async () => {
     const root = await makeFixture({
       'dist/component.svelte': `<script lang="ts">
   export const label = 'ok';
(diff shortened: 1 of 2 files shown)


stevekinney:
Round-2 findings (codex, copilot) — all six verified empirically, four real code fixes plus one refuted-but-still-improved test nit. Resolving all threads with this comment (the CSS-in-`<style>`-block finding was reported independently by both bots, same fix resolves both).

**1 & duplicate: CSS `@import` inside a `.svelte` `<style>` block was never scanned.** Confirmed by code inspection — the `.svelte` branch extracted and scanned only `<script>` blocks, then `continue`d unconditionally. Fixed with `extractSvelteStyleBlocks`, mirroring the existing `<script>` extraction, feeding its output through the same CSS-`@import` scan `.css` files already use.

**2: quote-aware script-tag parsing.** Reproduced first: `t.scanImports('">\n  export let value: T;\n')` (the malformed output the old `[^>]*` extraction produced for a real `generics="T extends Array<string>"` attribute) throws `"Unterminated string literal"` — confirmed this would crash the release validator on legitimate Editor source, not a contrived case. Fixed with `findTagEnd`, a small quote-tracking scan that finds the tag's real closing `>`.

**3: dynamic import with a computed specifier.** Reproduced: `t.scanImports("import('undeclared-package/' + feature);")` returns `[]` — confirmed the lexer alone is blind to this, and confirmed separately that the *old* regex did catch the literal prefix (it matches any `import(` followed immediately by a quote, regardless of what follows). Fixed with a narrow supplementary regex scoped only to `import(` + immediate string literal — not a reintroduction of the old `from`-matching branch, which was the actual source of the original bug.

**4: bare Node builtins via `require()`.** Reproduced: `t.scanImports("require('fs')")` → `{"kind":"require-call","path":"fs"}`, and the old filter only exempted the `node:`-prefixed spelling. Also confirmed `import('fs')` (dynamic, fully static) has the identical problem via `dynamic-import` kind — the fix isn't require-specific. Fixed via `node:module`'s own `builtinModules` (bare + `node:`-prefixed + subpaths), not a hand-maintained list.

**5 (test file): the claimed `mkdir(join(fullPath, '..'))` EISDIR failure.** Reproduced the exact scenario with `mkdir(..., { recursive: true })` (which the original code already had) against real nested and flat paths — it did not fail; `path.join`'s lexical normalization correctly cancels a trailing `/<segment>/..` regardless of whether `<segment>` looks like a file or a directory. I'm not disputing the finding as a code-quality point, though: switched to `dirname()` anyway, since it says what it means instead of relying on a normalization side effect a reader has to re-derive to trust, and left a comment noting the refutation so a later round doesn't rediscover this as a live bug.

**Verification**: every fix has a standalone repro against the real `Bun.Transpiler` run before the fix (see above), plus committed regression coverage in `validate-consumer.test.ts`. Break-and-restore: probed all four substantive fixes at once (kept the new exports so the test file still loads, reverted each function's internal logic to its pre-round-2 shape) — exactly the 9 tests pinning these specific fixes went red, the other 18 (including every round-1 finding's own coverage) stayed green regardless. Restored and hash-verified byte-identical, re-confirmed all 27 tests green.

Full suite: `bunx turbo run test typecheck lint --filter=@lostgradient/editor --force` (856 pass, 1 pre-existing skip, 0 fail, 0 typecheck/lint errors), and the real `bun run --filter=@lostgradient/editor validate:consumer` pipeline still completes successfully end-to-end.

stevekinney:
Round-3 findings (Cursor + codex x6) — all seven resolved: five real fixes (two of them via one governing change each), two refuted with dist-inventory evidence per the coordinator's principle that the actual published artifact, not a hypothetical enumeration, is the spec.

**1 (Cursor, "regex still scans comment import()") — removed, not patched.** Round 2's supplementary `import(` regex ran over the same raw, comment-bearing source text that caused #1334, and reopened the identical failure mode for `import(` specifically. Deleted entirely. The resulting gap (a computed dynamic import's literal prefix is no longer caught) is documented in the module's new header limitations section: verified via `npm pack --dry-run` + tarball extraction (the actual published artifact, not the raw workspace `dist/`, which still contains excluded `.test.js` files) that zero computed dynamic imports exist anywhere in it today, backed by the runtime fixture (`buildConsumerEntries`/`runPlainNodeConsumer` build and execute every export target with only declared peers installed).

**2 & 2 (codex x2, Bun-vs-Node builtins + subpath exact-membership) — one fix for both.** Reproduced first: under Bun, `require('node:module').isBuiltin('ws')` and `isBuiltin('undici')` both incorrectly return `true` (confirmed — Bun ships these as built-in shims; real Node does not), and real Node correctly returns `false` for both plus `false` for `assert/not-real` while correctly returning `true` for `fs/promises`. Fixed by spawning the actual target Node executable (reusing `runPlainNodeConsumer`'s existing candidate-search logic, extracted as `findRealNodeExecutable`) and asking it `module.isBuiltin(specifier)` directly — Node's own logic, not a hand-rolled first-segment heuristic, fixes both findings in the same motion.

**3 (codex, "quoted CSS paths with spaces") — real gap, fixed.** Renamed to `isPlausibleCssImportSpecifier` and changed to validate only the bare-package-name portion of a capture (`barePackageName(specifier)`), not the whole string — `'undeclared-package/theme dark.css'` now correctly passes (and is checked against declared deps) while comment prose (which has no `/`, so its own "bare name portion" is the whole whitespace-laden string) is still rejected. Stays fail-closed.

**4a (codex, "dynamic imports in Svelte template expressions") — refuted with dist evidence.** Grepped the actual published artifact for `import(` calls outside every `<script>` block (stripped script blocks first, then searched what remained): zero matches. Every `import(` call site in the real output is inside a `<script>` block with a fully static specifier. Documented as an accepted gap in the header (regexing markup would reopen the exact failure class this round exists to close), backed by the same runtime-fixture argument as finding 1.

**4b (codex, "locally shadowed require") — refuted with dist evidence, more strongly than expected.** Grepped the actual published artifact for `require(` call sites of any kind: zero. This package's output is 100% ESM. There's no `require()` usage at all for a local binding to shadow, today. Documented in the header as a structurally-absent pattern, revisit if that ever changes.

**4c (codex, "type-only imports in Svelte sources") — real, live pattern; documented per the explicit instruction.** Confirmed `import type {...}` is used extensively throughout this package's actual `.svelte` source (currently always against an already-declared peer or a relative import, so no live violation today). Checked whether `Bun.Transpiler` has a config option to surface type-only imports before concluding it can't be done: no such flag exists in the documented `TranspilerOptions` API, and empirically `scan`, `scanImports`, and `transformSync` all erase `import type` identically. Per the instruction ("if not, document it"), documented in the header rather than adding a new regex — noting the partial backstop: `runSvelteCheckConsumer`'s `svelte-check` pass currently only exercises `MarkdownEditor`'s own transitive type-import graph, not `ReviewEditor`'s or `DiffViewer`'s independently, named honestly rather than overclaimed as complete coverage.

**Regression coverage** (35 tests in `validate-consumer.test.ts`): the original #1334 prose case, the new comment-prose-with-`import()` case (finding 1), the real-Node-vs-Bun distinction for `ws`/`undici` and the `assert/not-real` exact-membership fix (finding 2), and the CSS whitespace-in-subpath fix (finding 3) are all pinned against the real functions.

**Verification**: reverting the whole file to the round-2 state produces a module-load error (the exported API surface itself changed with this round). A more surgical, targeted probe -- swapping only `nodeBuiltinMembership`'s real-Node spawn for Bun's own `node:module` -- makes exactly the 2 tests pinning the `ws`/`undici` distinction fail, all other 33 tests stay green regardless (confirming they don't depend on that specific fix). Restored and hash-verified byte-identical, re-confirmed all 35 green.

Full `bunx turbo run test typecheck lint --filter=@lostgradient/editor --force` (864 pass, 1 pre-existing skip, 0 fail, 0 typecheck/lint errors), and the real `bun run --filter=@lostgradient/editor validate:consumer` pipeline still completes successfully end-to-end.

Module header now carries a full limitations section naming every deliberate static-analysis gap and why the runtime fixture (or, for type-only imports, the partial svelte-check backstop) makes it acceptable rather than a hole.

PR #1471, opened 2026-08-29
Add terminal, data, guidance, and utility components

stevekinney:
## Summary

- add RelativeTime, Citation, CodeLocation, QuotaMeter, and Badge severity support
- add TerminalOutput, TerminalFrame, FindBar, ZoomPanViewer, KeyValueEditor, DonutChart, and resizable DataTable columns
- add ModalRegion/useModal, GuidanceRegion/useGuidance, and InlineConfirm
- regenerate component metadata, exports, examples, styles, and token documentation

## Validation

- `bun test --conditions browser --conditions svelte --parallel=1 src/components/{badge,citation,code-location,data-table,donut-chart,find-bar,guidance-region,inline-confirm,key-value-editor,modal-region,quota-meter,relative-time,setting-row,terminal-frame,terminal-output,zoom-pan-viewer}/*.test.ts`
- `bun test --conditions browser --conditions svelte --parallel=1 packages/components/src/styles/check-token-contrast.test.ts`
- `bun run --filter=@lostgradient/cinder components:check`
- `bun run --filter=@lostgradient/cinder exports:check`
- `bun run --filter=@lostgradient/cinder typecheck`
- `bun run --filter=@lostgradient/cinder lint`

Closes CIN-375, CIN-298, CIN-297, CIN-287, CIN-286, CIN-283, CIN-282, CIN-281, CIN-280, CIN-279, CIN-278, CIN-275, and CIN-274.

PR #1477, opened 2026-08-31
Add the missing README Usage sections for fifteen components

stevekinney:
## Summary

Fifteen components added across #1469, #1471, and #1472 shipped READMEs with a title and a description but no `## Usage` section. The README compile gate reports `no-heading` for each, which fails `validate:consumer` at its `typescript-consumer` step — blocking **both** the release and `main-green` (which runs `validate:consumer:readme-usage-examples`).

`donut-chart`, `find-bar`, `form`, `guidance-region`, `host-provider`, `inline-confirm`, `key-value-editor`, `modal-region`, `parameter-field`, `policy-lock`, `setting-row`, `shortcut-field`, `terminal-frame`, `terminal-output`, `zoom-pan-viewer`.

## These examples are not invented

Each of the fifteen already had exactly one authored example in its `<component>.examples.json`. Despite its name, `generate-readme-usage-examples.mjs` only *validates* the README fence — it never writes it. So these sections are transcribed from the existing authored examples rather than newly written.

That matters for review: the usage shown is the component author's, and the README now agrees with the examples corpus instead of being a second, independently-drifting source of truth.

## Verification

```
bun run --filter=@lostgradient/cinder validate:consumer:readme-usage-examples
```

Exit code `0`, against the real packed tarball:

```
[validate-consumers] typescript-consumer OK (gates 1–3 green).
[validate-consumers] readme usage examples smoke passed.
```

Every fence is genuinely compile-checked, not merely present — the gate materializes each one as a standalone `.svelte` file that the existing `svelte-check --threshold error` pass picks up.

## No changeset

Component READMEs *do* ship (`src/components/**/*.md` is in `files`), so this would normally warrant one. It doesn't here: all fifteen components arrived after the release stalled on 2026-08-21 and are absent from the published `0.24.8`, so this documentation has never shipped and lands with the components themselves in `0.25.0`.

## Release context

This is the last of four independent blockers between `main` and a published release, found one behind the other:

1. InlineConfirm viewport `@media` — #1473, merged
2. Chat `--cinder-ease-out` undeclared token — #1474
3. `tokens-consumer` stale non-representable assertion — #1476
4. these missing `## Usage` sections — this PR

The common cause is that #1469/#1471/#1472 landed large component batches while the release PR sat unmerged since 08-21, so no real publish ever exercised these gates.

https://claude.ai/code/session_01WbEcbbzejHmM77KousLXYy

### DIFF
--- a/packages/components/src/components/donut-chart/README.md
+++ b/packages/components/src/components/donut-chart/README.md
@@ -12,3 +12,19 @@ Use a bar chart when precise comparison matters more than the part-to-whole rela
   onSeriesClick={(datum) => selectStatus(datum)}
 />
 ```
+
+## Usage
+
+```svelte
+<script lang="ts">
+  import { DonutChart } from '@lostgradient/cinder/donut-chart';
+  const data = [
+    { label: 'Build', value: 42 },
+    { label: 'Review', value: 28 },
+    { label: 'Test', value: 18 },
+    { label: 'Docs', value: 12 },
+  ];
+</script>
+
+<DonutChart label="Workload" {data} centerLabel="Tasks" valueLabels />
+```
--- a/packages/components/src/components/find-bar/README.md
+++ b/packages/components/src/components/find-bar/README.md
@@ -1,3 +1,13 @@
 # FindBar
 
 Docked find-in-document controls for a host-provided search backend.
+
+## Usage
+
+```svelte
+<script lang="ts">
+  import { FindBar } from '@lostgradient/cinder/find-bar';
+</script>
+
+<FindBar value="terminal" activeIndex={1} matchCount={4} />
+```
--- a/packages/components/src/components/form/README.md
+++ b/packages/components/src/components/form/README.md
@@ -3,3 +3,23 @@
 Form is a thin native form root that exposes asynchronous submission state to its child snippet and ignores duplicate submissions while the current handler is pending.
 
 Use native validation and form controls as usual. The component owns only submit coordination; it does not replace native serialization, validation, or reset behavior.
+
+## Usage
+
+```svelte
+<script lang="ts">
+  import { Form } from '@lostgradient/cinder/form';
+  import { Input } from '@lostgradient/cinder/input';
+  import { Button } from '@lostgradient/cinder/button';
+</script>
+
+<Form onSubmit={() => undefined}>
+  {#snippet children({ submitting })}
+    <Input id="form-name" label="Name" value="" disabled={submitting}>
+      {#snippet leading()}{/snippet}
+      {#snippet trailing()}{/snippet}
+    </Input>
+    <Button type="submit" disabled={submitting}>Save</Button>
+  {/snippet}
+</Form>
+```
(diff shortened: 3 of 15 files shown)


stevekinney:
## Not ready to merge — the review feedback is correct, and my framing above was wrong

@chatgpt-codex-connector raised seven P2 objections and I think every one of them lands. Recording that plainly rather than arguing or quietly merging a green-able gate.

My PR description claims these examples are safe because they were "transcribed from the component author's own `examples.json` rather than newly invented." That is true about their provenance and misleading about their fitness. `examples.json` entries are minimal fixtures; a README `## Usage` block is teaching material. Those are different artifacts with different bars, and moving the text between them without raising it to the second bar is what produced these defects:

- **InlineConfirm** — the example sets `open` with no trigger, so the component captures whatever happened to be focused at mount and has no trigger to restore focus to. That directly contradicts the sentence immediately above it in the same README ("restores focus to the trigger when dismissed").
- **TerminalFrame** — wraps static read-only output, contradicting the repo's own guidance to reach for `TerminalOutput` in that case, and demonstrates none of the resize contract.
- **FindBar** — omits `onQueryChange`, so typing updates local state and then does nothing; the navigation controls never render.
- **ModalRegion** — discards the `Promise<boolean>` from `modal.confirm()`, so confirming and cancelling are indistinguishable and the named action never runs. The promise *is* the component's distinguishing API.
- **Form** and **SettingRow** — empty adornment snippets are still truthy, so `Input` renders its grouped-control frame with blank padded spans.
- **GuidanceRegion** — trigger ARIA lands on a non-focusable `<span>` instead of the nested button.

Fixing these properly means authoring seven-plus genuine usage examples against each component's intended semantics, then re-running the compile gate. That is real documentation work on components I did not write, and I would rather hand it over than guess a second time — I already guessed once, in the direction of "this is mechanical," and this review is the evidence that it wasn't.

**Leaving this PR open and unmerged.** Merging it would turn the gate green while shipping misleading docs in a published package, which is worse than the red gate it fixes.

### CI status, for completeness

`playwright-visual shard 7` failed here on `domain focus rings -- chat harness`. That is a flake, not this branch: the identical shard **passed** on #1476, which was rebased onto `b3e28ace` — a tree that already contains the chat easing change from #1474 — and README prose cannot affect focus-ring painting. Worth a re-run rather than a fix.

### What is and isn't unblocked

The other three release blockers are merged: #1473, #1474, #1476. This gate is the last one standing between `main` and a publish, and it blocks `main-green` as well as `release` (`main-green` on `b3e28ace` now gets past the chat token audits and stops exactly here).

Note also that `@lostgradient/cinder@0.25.0` and `@lostgradient/chat@0.13.0` are versioned in `package.json` on `main` but **not published** — the #1406 release run aborted at the token-consumer gate before publishing. Once this lands, `changeset publish` should pick them up, since their versions are absent from the registry.

stevekinney:
Superseded by #1478, which covers these same fifteen READMEs plus the packed Chat composer fix and a changeset. The review feedback from this PR has been carried over to #1478, where the usage-example content is byte-identical and the same seven objections apply.

PR #1478, opened 2026-08-31
Fix package release validation

stevekinney:
## Summary

- distinguish registry tokens that can be represented in resolved-context JSON
- make packed Chat composer sources parse under the minimum Svelte/Vite consumer
- add compiling Usage examples for newly added component READMEs

## Validation

- `bun run --filter=@lostgradient/cinder tokens:check`
- `bun test packages/components/fixtures/typescript-consumer/generate-readme-usage-examples.test.ts packages/components/scripts/tokens/registry.test.ts`
- `bun run --filter=@lostgradient/chat typecheck`
- `bun run --filter=@lostgradient/cinder validate:consumer`


stevekinney:
## Closing #1477 in favor of this — but its review feedback applies here verbatim

I opened #1477 against the same fifteen READMEs before seeing this. This PR is a superset (it also carries the packed Chat composer fix and a changeset), so I've closed mine. One thing is worth carrying across rather than losing with it.

**The two PRs produced byte-identical usage examples** — both transcribed each component's existing `examples.json` entry. On #1477, @chatgpt-codex-connector reviewed that content and raised seven P2 objections. Because the content here is identical, every one of them applies to this PR unchanged:

- **InlineConfirm** — sets `open` with no trigger, so the component captures whatever was focused at mount and has no trigger to restore focus to. This contradicts the sentence directly above it in the same README ("restores focus to the trigger when dismissed").
- **TerminalFrame** — wraps static read-only output, contradicting the guidance to use `TerminalOutput` for read-only streams, and demonstrates none of the resize contract.
- **FindBar** — omits `onQueryChange`, so typing updates local state and then does nothing; navigation controls never render because neither callback is supplied.
- **ModalRegion** — discards the `Promise<boolean>` from `modal.confirm()`, so confirm and cancel are indistinguishable and the named action never runs. That promise is the component's distinguishing API.
- **Form** / **SettingRow** — empty adornment snippets are still truthy, so `Input` renders its grouped-control frame with blank padded spans.
- **GuidanceRegion** — trigger ARIA (`aria-expanded`/`aria-haspopup`/`aria-controls`) lands on a non-focusable `<span>` rather than the nested button.

The underlying issue is that `examples.json` entries are minimal fixtures while a README `## Usage` block is teaching material — different artifacts, different bars. The compile gate only checks that the fence parses and renders the component's tag, so transcription satisfies CI while still shipping examples that mislead. Worth a decision on whether to raise these to teaching quality now or land them and follow up, since these READMEs **do** ship (`src/components/**/*.md` is in `files`).

### Two notes on the release state

- The tokens/resolved-context half of this PR's description is **already on `main`** via #1476, which taught the consumer fixture about the four `nonRepresentableValue` tokens. Worth confirming this branch doesn't re-litigate it — the current diff looks clean of it.
- `@lostgradient/cinder@0.25.0` and `@lostgradient/chat@0.13.0` are versioned in `package.json` on `main` but **not published**: the #1406 release run aborted at the token-consumer gate. Once this lands and `main-green` is green, `changeset publish` should pick both up, since those versions are absent from the registry. Registry currently serves 0.24.8 and 0.12.0.

### Flake worth knowing about

`playwright-visual shard 7` (`domain focus rings -- chat harness`) failed once on #1477. It is a flake, not a real regression: the identical shard passed on #1476 against a tree that already contained the chat easing change, and README prose cannot affect focus-ring painting.

PR #1496, opened 2026-09-01
feat(virtual-list): dynamicSize measurement, typed scroll ref, and the CIN-204 guard

stevekinney:
Wave 1 of [Cinder: Virtual list improvements](https://linear.app/lost-gradient/project/cinder-virtual-list-improvements-d2df7b27d14c/overview). Adds opt-in measured variable-height rows, a typed imperative handle, and the CI guard that keeps the engine dependency-free.

Closes CIN-186, CIN-187, CIN-188, CIN-189, CIN-190, CIN-202, CIN-204, CIN-205.

## What this adds

`dynamicSize` measures each rendered row with a single component-owned `ResizeObserver` and caches the result by key, using `itemHeight` as the estimate for rows not yet measured. `bind:ref` exposes a typed `VirtualListRef` with `scrollToIndex(index, options)`, accepting `align` and `behavior`.

## The two decisions worth reviewing

**Anchor resolution uses a pre-mutation snapshot.** `offsets` is `$derived` off the same version counter `record()` bumps, so by the time the correction runs, the table has *already* rebuilt to include the very measurement being corrected for. Binary-searching it finds the wrong anchor. A `previousOffsets` snapshot is threaded across effect runs and the anchor is always resolved against that. This was caught by an adversarial judge against the original design, not by a test.

**The correction is absolute, written in a paired effect.** `$effect.pre` computes an absolute target from pre-patch state; a plain `$effect` writes it after Svelte patches the DOM, still before paint. A relative `scrollTop += delta` written pre-patch gets clamped by the browser against the stale content size and partially lost. That split is also how CIN-202 is satisfied — Svelte's own scheduler rather than a hand-rolled double-rAF — and a test asserts several measurements in one flush produce **exactly one** scroll write.

A smaller but load-bearing detail: rows carry **no inline height** under `dynamicSize`. A pinned height would make every row measure back as exactly the estimate, so the measurement pass could never observe a real size. Both the SSR suite and a component test pin this.

## CIN-190: the fixed path is genuinely untouched

No measurement, no caching, no correction, and no `ResizeObserver` observing any row. A test drives a full render, scroll, and append cycle and asserts no row is ever observed.

The assertion is deliberately scoped to *rows* rather than to observer construction as a whole, because the component has always observed its own scroll container for viewport size. Asserting zero constructions would be asserting something untrue about pre-existing behavior.

## CIN-204

`check:virtual-list-dependency-free`, wired into `lint:invariants` (which runs in `unit-tests` and `main-green`) and declared in `check-pipeline-coverage`'s table.

It scans static imports for `@tanstack/virtual-core` and for undeclared bare specifiers, and scans dynamic `import()` for the forbidden specifier **only**. That asymmetry is deliberate: test files in this subtree legitimately dynamic-import devDependencies (`await import('@testing-library/svelte')`), so applying the full rule there would be all false positives, while the narrow rule still closes the evasion. A companion Bun test asserts the same invariant independently, so a local `bun test` catches a violation without CI.

## Verification

- 121 tests pass across 6 files. Typecheck 0 errors. Lint 0 errors, none in these files. `components:check` clean.
- `measurement-window.ts` and `frame-scheduler.ts` at **100% lines and functions**.
- Engine math was differentially fuzzed against independent brute-force reference implementations — **92,508 assertions**. `findOffsetIndex` matches a linear scan across randomized tables *including zero-size items*, which produce duplicate offsets, exactly where a binary search classically picks the wrong equal. `getDynamicVirtualWindow` was checked against the invariant that every item overlapping the viewport is mounted.
- `data-table`, the other consumer of `fixed-virtual-window.ts`, is green. That file is not touched at all.

## Notes for the reviewer

- Wave 1 of 5. The engine signatures are intended to survive Waves 2-4 unchanged; horizontal/RTL, reverse/infinite/window-scroll/restoration, and sticky/keyboard/adaptive-overscan land next.
- The stale "v1 requires a fixed `itemHeight`" claim is removed from the README and `@avoidWhen` now rather than deferred to the docs wave, since shipping `dynamicSize` makes it false at this merge.
- The pre-commit hook reported 12.7s against a 10s budget. It is a warn-never-gate signal and the cause is commit size (23 staged files through prettier), not a hook regression, so the budget is deliberately left alone rather than raised to absorb one large commit.

https://claude.ai/code/session_01BicogfzUyxNrcZZumwGJSp


stevekinney:
## Review round 1 addressed — all 16 threads

Every finding from Copilot and Codex was legitimate. Nothing was waved off. Grouped by what they actually were, with the reasoning where it matters.

### Real defects

**Rounded measurements compounded error down the offsets table.** (Codex P1) `Math.round` on every measurement meant 5,000 rows at 20.4px cached as 20px left the spacer ~2,000px short, and every `scrollToIndex` target past the top of the list wrong. Sizes are cached at full precision now; an epsilon is used *only* for change detection, which is what actually breaks the correction → relayout → same-size-reported → correction loop. Since sizes are stored exactly, a re-report is normally bit-identical and would be caught even at zero — the epsilon is margin against device-pixel-ratio jitter, not a substitute for exact storage. A regression test sums 5,000 fractional measurements and asserts no drift.

**`stickToBottom` lost the bottom when an appended row measured taller.** (Codex P1) The append pin scrolls to the total *as estimated*. A row that then measures taller grows the total without an item-count change, and the anchor correction ignores it by design because it sits below the anchor — so the viewport sat short of the bottom, contradicting what the prop promises. It re-pins on total-size growth now.

**A row measuring zero was discarded.** Zero is a legitimate measurement; rejecting it left the table reserving space the row no longer occupied. Only non-finite and negative are rejected.

**The cache was never pruned.** A long-lived feed that filters or rolls over its contents kept every size ever measured, so memory tracked history rather than the current collection. `prune()` is wired to the key set now.

**A mode switch during an append yanked the reader to the end.** `previousDynamicTotalSize` stays 0 while fixed mode runs, so switching to dynamic mid-append made `isAtBottom(…, 0, …)` true for *any* offset. Guarded on the mode being unchanged.

**Overlapping `scrollToIndex` calls fought each other**, and the older loop could land last. Each call supersedes the previous via a generation token.

**The settle loop waited exactly one frame per attempt**, so a smooth scroll spent all three attempts mid-flight and gave up before the destination rows mounted. It waits for the scroll position to hold still now, frame-capped so a continuously-scrolling container can't hold it open.

**The row `ResizeObserver` was resolved from the outer `globalThis`.** A list mounted in an iframe got a constructor bound to the wrong document, or none, leaving every row unmeasured. Resolved from the node's own document now, matching `useResizeObserver`.

**The dependency guard exempted all dynamic imports** from the undeclared-bare-import rule. Shipped source faces the full rule now; only test files keep the exemption, since nothing in them ships. The `@tanstack/virtual-core` ban still applies to both, and there are tests for each side of that asymmetry.

### Documentation that was simply wrong

Six threads flagged the same overstatement: the prop JSDoc, README, schema, and changeset all claimed *no* `ResizeObserver` is constructed anywhere when `dynamicSize` is off. The component has always observed its own scroll container for viewport size. Copilot was right, and my own test had already scoped its assertion to *rows* for exactly this reason — the prose contradicted the test. Narrowed everywhere and regenerated.

`behavior` is now the DOM `ScrollBehavior` type rather than a hand-rolled union.

### Two defects the reviewers did not find, surfaced by writing the tests for the ones they did

**The re-pin effect never re-ran.** It guarded on `isPinnedToBottom` *before* reading `offsets`, so on the run where the pin wasn't yet armed it never registered the total size as a dependency. My own test caught this — the fix looked right and was inert. Both reactive reads now happen before any early return.

**The teardown effect was keyed on `dynamicSize`.** A parent re-rendering with an *unchanged* `dynamicSize` still re-ran it, and its cleanup disconnected the shared row observer while the row attachments — unchanged — never re-observed. Every row went silently unmeasured with the component still looking healthy. Found by bisecting the re-pin failure and tracing observer counts through render → scroll → rerender. Teardown is now a single dependency-free effect that runs once, at destroy.

That second one is the more serious of the two: it would have shipped as "dynamic sizing silently stops working after a parent re-render."

### Verification

201 tests pass across virtual-list and data-table. Typecheck 0 errors. Lint 0 errors, none in these files. `components:check`, `check:virtual-list-dependency-free`, `check:pipeline-coverage`, and `check:test-cleanup` all clean.

https://claude.ai/code/session_01BicogfzUyxNrcZZumwGJSp

